PR #59, opened 2022-01-28
Introduce ActionPointsAdjustedRule.

orendain:
This rule adjusts the starting action points for any specified PieceConfig.

@jimconner WDYT?

---

FWIW, results of testing:

My argument looked something like:
```
new Dictionary<string, int> {
{ "HeroGuardian", 1 },
{ "HeroSorcerer", +6 },
{ "HeroHunter", -1 },

{ "Rat", 5 },
{ "Spiderling", 0 },
{ "RatKing", -2 },
};
```

Despite additional APs, the monsters I tested did not attack multiple times.  They did, however, used their extra AP to close in from across the map.  I imagine they must be hardcoded/configured to perform their ability only a set number of times, but can use up their AP for generic actions like movement (although I haven't looked at the code to verify).

With a net AP of zero (e.g., `RatKing -2`), I confirm the monster did not get to move or attack.


jimconner (approved):
LGTM

PR #69, opened 2022-01-29
Introduce generic ability damage rule.

orendain:
Similar to other proposed generic rule: https://github.com/orendain/DemeoMods/pull/59

@jimconner Any thoughts as to how the corresponding crit should be adjusted?


jimconner:
In general crit damange seems to be about 75%-100% more than standard damage. I think that setting a standard where critical damage is twice normal isn't too unreasnable. It would be easy enough to pass in two ints for standard and crit damage, but I can't really envisage many occasions where people would want to set that differently.

Rules/Rule/AbilityDamageAdjustedRule.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        protected override void OnPostGameCreated()
+        {
+            var abilities = Resources.FindObjectsOfTypeAll<Ability>();
+            foreach (var item in _adjustments)
+            {
+                var ability = abilities.First(c => c.name.Equals($"{item.Key}(Clone)"));
+                ability.abilityDamage.targetDamage += item.Value;
+            }
```
jimconner, line 31: I think we should go with crit = 2x standard damage before we merge this... Otherwise it'll be confusing when folks roll double swords and do less damage than normal.

jimconner (changes requested):

Rules/Rule/AbilityDamageAdjustedRule.cs
orendain: Good idea - done.

jimconner (approved):

jimconner (approved):

PR #73, opened 2022-01-29
Add a rule for adjusting character's default MoveRange

jimconner:
Another quick PR for adjusting the Movement Range of characters (same as the speed potion does).

Appears to all work OK.

### DIFF
--- a/Rules/README.md
+++ b/Rules/README.md
@@ -11,6 +11,7 @@ RulesAPI.
   of a RulesAPI rule.
 - **AbilityDamageAdjustedRule**: Ability damage is adjusted
 - **ActionPointsAdjustedRule**: Action points are adjusted
+- **MoveRangeAdjustedRule**: Movement Range(s) are adjusted
 - **RatNestsSpawnGoldRule**: Rat nests spawn gold
 - **StartHealthAdjustedRule**: Starting Health is adjusted
 - **ZapStartingInventoryAdjustedRule**: Zap starting inventory is adjusted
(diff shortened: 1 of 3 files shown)


jimconner:
On a side-note to this... ActionPointsAdjusted, StartHealthAdjusted and MovementRangeAdjusted are all nearly identical bits of code, apart from the name of the thing that is being adjusted. Should we think about making a generic PiecConfigAdjuster which can take multiple params... We could then define in a single rule... `new Rule.PieceConfigAdjustedRule(new Dictionary<string, string, int>() {{ "HeroSorcerer", "ActionPoints", 5 }, {"HeroGuardian", "MoveRange", 10}, {"HeroHunter","StartHealth", 20}, etc`

jimconner:
As an aside... The code for ActionPointsAdjusted, StartHealthAdjusted and MoveRangeAdjusted is nearly identical apart from the name of the field being changed. Should we think about making a PiceConfigAdjusted rule which can take the name of the field to be adjusted as a parameter too. The ruleset could then be something along the lines of...

```
new Rule.PieceonfigAdjustedRule(new SomeDataType<string, string, int>() { 
                                                                                                          { "HeroSorcerer", "MoveRange", 20 },
                                                                                                          { "HeroGuardian", "ActionPoints", 2 },
                                                                                                          { "HeroHunter", "StartHealth", 20 },
                                                                                                          { "WolfCompanion", "StartHealth", 20 },
                                                                                                          { "HeroBard", "VisionRange", 40 },
                                                                                                          { "RatKing", "MoveRange", -10},
                                                                                                         }),
```

jimconner:
This PR is superceeded by #74, so I don't think it's worth merging this one in. I'm going to close it.

PR #74, opened 2022-01-29
A generic PiceConfig Adjuster Rule which can set any named integer field.

jimconner:
# What
We had nearly identical code for ActionPointsAdjusted, StartHealthAdjusted and MoveRangeAdjusted, The only difference between them was the name of the parameter which they were adjusting.

I thought I'd have a go at making a totally generic PiceConfigaAdjusted rule which take a lists of PieceName, ParameterName and the value to set it to.

Because I don't know C# and am learning as I go, I've used `List<List<string>>` as the datatype to pass aroud. This is not ideal as we want to set int values - I used `int.Parse()` within the rule to convert the numeric string value to int. There will certianly be room for improvement here.

Aside from the code quality issues, this appears to function correctly.
  
## Suggested Ruleset for Testing

```
var sorcererTestRules = new HashSet<RulesAPI.Rule> {
                                            new Rule.ZapStartingInventoryAdjustedRule(),
                                            new Rule.AbilityDamageAdjustedRule(new Dictionary<string, int>() { { "Zap", 8 } }),
                                            new Rule.PieceConfigAdjustedRule(new List<List<string>>() {new List<string>(){"HeroSorcerer","MoveRange", "10"}, 
                                                                                                       new List<string>(){"HeroSorcerer","ActionPoint","3"},
                                                                                                       new List<string>(){"HeroSorcerer","StartHealth", "20"}
                                                                                                        } ), 
                                          };
var sorcererTestRuleset = RulesAPI.Ruleset.NewInstance("SorcererTest", "Test set of rules which apply to the Sorcerer.", sorcererTestRules);
registrar.Register(sorcererTestRuleset);
```

### DIFF
--- a/Rules/Rule/PieceConfigAdjustedRule.cs
+++ b/Rules/Rule/PieceConfigAdjustedRule.cs
@@ -1,0 +1,47 @@
+﻿namespace Rules.Rule
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using Boardgame;
+    using HarmonyLib;
+    using UnityEngine;
+
+    public sealed class PieceConfigAdjustedRule : RulesAPI.Rule
+    {
+        public override string Description => "Start Health is adjusted";
+
+        private readonly List<List<string>> _adjustments;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PieceConfigAdjustedRule"/> class.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


orendain (approved):
Awesome thinking!

Looks like most flat `PieceConfig` fields are ints anyway, with the exception of a few floats and two booleans. 👍🏾

Added this doc for reference: https://github.com/orendain/DemeoMods/blob/main/docs/PieceConfig.md

PR #75, opened 2022-01-29
Missed commit from StartHealthAdjustment PR

jimconner:
I thought I'd included this in the last PR after you approved, and before I merged, but apparetly I didn't.


jimconner (commented):
Line endings have been changed by my editor, which is why I didn't have this in the PR in the first place. Not sure how to stop VisualStudio from doing this.

jimconner:
This same fix is included in #74 already, so if you like and merge that one, you won't need this one.

orendain:
> This same fix is included in #74 already, so if you like and merge that one, you won't need this one.

Done!  Closing.

PR #82, opened 2022-01-30
actionPoints varialbe name was out of context.

jimconner:
Changed to 'property' as we're setting a property within a PieceConfig.

No functionality changes... Just tidying up.


orendain (approved):
Awesome.

PR #89, opened 2022-01-31
AbilityActionCostRule - A rule for adjusting whether specific cards have a cost to cast or not.

jimconner:
# What

As a quality of life improvement, it can be nice to set some of the cards which usually have a cost to be free to cast instead. In particular I feel that the Wizard is a much better character when Zap is a 0AP cost. It might also work well for the Bard's Courage buff which is usually in high-demand and effectively means that the Bard only has one action per turn for themselves. 

This new rule takes a Dictionary<string, bool> containing the names of the cards, and the value you'd like them to be set to. I originally thought about just passing a list of strings, but realised there might be other use cases (such as if you'd boosted Hero Pieces to have more AP, but wanted to make Heal have a cost to cast etc).

I was a little bit lazy around the OnDeactivate method, and just set the values to the inverse of whatever was specified in the Dict. If someone were to pass in a Dict with the values unchanged from the defaults, then it would mean that when Deactivated they would change to non-default values. I'm not sure if this would cause any real-world issues, but will happily invest some more time solving it if you think it's worthwhile. 

## Suggested Ruleset for reviewing this PR.
```
var sorcererTestRules = new HashSet<RulesAPI.Rule> {
                                new Rule.ZapStartingInventoryAdjustedRule(),
                                new Rule.AbilityDamageAdjustedRule(new Dictionary<string, int>() { { "Zap", 8 } }),
                                new Rule.AbilityActionCostAdjustedRule(new Dictionary<string, bool>() { { "Zap", false } }),
                                };
var sorcererTestRuleset = RulesAPI.Ruleset.NewInstance("SorcererTest", "Test set of rules which apply to the Sorcerer.", sorcererTestRules);
registrar.Register(sorcererTestRuleset);
```

### DIFF
--- a/Rules/Rules.csproj
+++ b/Rules/Rules.csproj
@@ -1,79 +1,80 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
-    <PropertyGroup>
-        <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-        <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-        <ProjectGuid>{8FC09405-8E06-4090-BB68-94882F0A52D8}</ProjectGuid>
-        <OutputType>Library</OutputType>
-        <AppDesignerFolder>Properties</AppDesignerFolder>
-        <RootNamespace>Rules</RootNamespace>
-        <AssemblyName>Rules</AssemblyName>
-        <TargetFrameworkVersion>v4.8</TargetFrameworkVersion>
-        <FileAlignment>512</FileAlignment>
-    </PropertyGroup>
-    <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-        <PlatformTarget>AnyCPU</PlatformTarget>
-        <DebugSymbols>true</DebugSymbols>
-        <DebugType>full</DebugType>
-        <Optimize>false</Optimize>
-        <OutputPath>bin\Debug\</OutputPath>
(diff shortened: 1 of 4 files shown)


Rules/Rules.csproj
```diff
@@ -1,79 +1,80 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <Content Include="README.md" />
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <Target Name="CopyToModsDir" AfterTargets="Build">
+    <Copy SourceFiles="$(OutputPath)\$(ProjectName).dll" DestinationFolder="$(SteamDemeoDir)\Mods" />
+    <Copy SourceFiles="$(OutputPath)\$(ProjectName).dll" DestinationFolder="$(OculusDemeoDir)\Mods" />
+  </Target>
+</Project>
```
orendain, line 80: https://stackoverflow.com/a/38556923  😅

orendain (approved):
LGTM!  Included a link on hopefully fixing indents 🙂 , but not urgent 👍🏾 .

Rules/Rules.csproj
jimconner: The Rules.csprod isn't being edited by me... It's happening automatically inside Visual Studio (not VSCode). I use VSCode for my day-to-day stuff, but I'm using VS(non-code) for the demo modding stuff because I don't know any better. Any idea how to stop Visual Studio from getting it wrong?

orendain: I use JetBrains Rider, but it looks like these are the steps that should work for VS:
1. Open the `.csproj` file: https://stackoverflow.com/a/64428453
2. Bottom-right of the screen, change indents: https://stackoverflow.com/a/38556923

Not urgent, of course 👍🏾

jimconner: The first step was OK... the second step sadly not. Visual Studio has the tabs/spaces settings buried in an options menu, and every one of them is set to 4-spaces and 'insert spaces'. They are different settings for different languages/filetypes and the setting for XML is already on 4-spaces - there is no filetype listed for .csprod files. 

Found a couple of posts along the same lines as the following one. No clear answers though.
https://stackoverflow.com/questions/68015170/visual-studio-web-csproj-file-indenting-flipping-between-tabs-and-spaces.

I manually fixed the .csproj for the AOE PR, and was about to fix this one too, but you merged it before I got the chance.

Issue #83, opened 2022-01-30
Consider sepearating rule types into static and non-static.


orendain:
Low-priority.

**Issue:**
- Due of limitations on how patching can be done (requires a static patch method, thus static field variables), it makes sense that only one instance any type of rule exists (multiple instances will only operate on the same static fields).
- However, this limitation means that even rules without patching requirements (e.g., `OnPatch()`) are limited to a single instance per class.

There are several reasons for why it would be beneficial to have multiple instances per rule type:
1. Increased granularity.  Instead of a single large PieceConfig rule, can have multiple PieceConfig rules (e.g., 1 per piece type).'
2. As a result of more granular (although still generic) rules, printing rule descriptions can incorporate value changes rather than just generic descriptions.

**Proposal:**
Splitting rules into static and non-static would look like the following:
- Rules with patch requirements (e..g, `Onpatch()`) would be one rule sub type, and those without would be another sub type.

jimconner:
"printing rule descriptions can incorporate value changes rather than just generic descriptions." - I was wondering if there was a way to parameterise these descriptions using the values being passed to generate the description text dynamically.

"PieceConfig(s) Adjusted: Sorcerer StartHealth +5, Guardian ActionPoints +1, WolfCompanion AttackDamage +3" etc

Python classes have an __init__ function which is great for this kind of thing, but I lack the necessary C# skills to know if the same can be achieved here.

orendain:
> "printing rule descriptions can incorporate value changes rather than just generic descriptions." - I was wondering if there was a way to parameterise these descriptions using the values being passed to generate the description text dynamically.

For sure!  The non-parameterized nature of the descriptions were mostly a function of not knowing exactly how the descriptions would be used (e.g., for printing on screen, or only used to describe the rule to users rule pre-initialization, etc.).

👍🏾

Issue #80, opened 2022-01-30
Add rule to set start cards.


orendain:
Target input could be something like:

```
- HeroClass: string
- Array of cards to grant
  - cardname: string
  - isReplenishable: bool
```

The code to patch doesn't deal in strings, but rather enums (i.e., integers).

It's trivial to map hero classes to their respective enum, given there's only 4 of them.  But with so many abilities/cards, I propose to just take ints as input.

Resulting in:

```
- HeroClass: string
- Array of cards to grant
  - abilityKey: int
  - isReplenishable: bool
```

jimconner:
Would we want this rule to also allow setting the number of cards which can be held? Given that Hunter can hold 11, there must bve a parameter somewhere that we can change.

orendain:
So it turns out the game is configured to allow a max of 2 replenishable cards.

Replenishable cards add +1 each to the max card count (9 + that count).

I remember playing around w/ setting max hand size a long time ago, but can't recall if that was possible to control host-side (there might be some hardcoded client-side logic there).  I'll report back what I find.

PR #92, opened 2022-02-01
Add rule for setting Sorcerer start cards.

orendain:
Part of https://github.com/orendain/DemeoMods/issues/80

This rule sets the starting cards for the Sorcerer.

**A quick copy/paste for testing:**
```
var cards = new List<Rule.SorcererStartCardsModifiedRule.Card>()
{
    new Rule.SorcererStartCardsModifiedRule.Card { Name = AbilityKey.Whirlwind, IsReplenishable = true },
    new Rule.SorcererStartCardsModifiedRule.Card { Name = AbilityKey.Freeze, IsReplenishable = false },
};

var sampleRules = new HashSet<RulesAPI.Rule> { new Rule.SorcererStartCardsModifiedRule(cards) };
```

### Issues with using a `Dictionary` or `List` or other:

I started by implementing the rule as generic, taking in a a `Dictionary<String, Dictionary<String, bool>>`.  That is, a mapping from hero name, to a set of (abilityKey, isReplenishable) values.

The biggest drawback with that is that dictionaries can not contain duplicate keys (e.g., can not specify two of the same cards).  Sure, there are workarounds, like adding a `count` field, but this felt too hacky.

Alternatively, I considered ` Dictionary<List<List<String>>>`, with custom parsing.  However, I didn't like the added complexity.

### So ... I eventually landed on one rule per hero:

Sure, this results in some duplicated code, but configuring each rule becomes easier by an entire order of magnitude.  This alone may not have been reason enough, except that I had also considered ...

### Future feature: Reading rule configuration from a file

Nested structures with a depth of 2+ are more prone to user confusion.  Additionally, it seems considerably easier (and neater) for a user to list starting cards for each hero as a single array.

Consideration for eventually writing to a file was also the reason why I opted to use a `struct`.  Not only do we get static typing out of the box with `struct`s with no parsing, but it turns out that `struct`s can be nicely [de/serialized using one of the libraries that ship with MelonLoader](https://github.com/LavaGang/MelonLoader/blob/5165b9ed9b441ec8c3d8f55f099edba609767f22/MelonLoader/TinyJSON/README.md).

We don't necessarily need to go with JSON.  We could abstract the config/parsing and see what all makes sense.

### Not necessarily a standard

I still believe rules should be implemented in whatever way it makes sense for the author, and that we will naturally trend towards improvement.


Rules/Rules.csproj
```diff
@@ -45,9 +45,10 @@
     <Compile Include="Rule\EnemyHealthScaledRule.cs" />
     <Compile Include="Rule\EnemyRespawnDisabledRule.cs" />
     <Compile Include="Rule\GoldPickedUpMultipliedRule.cs" />
+    <Compile Include="Rule\PieceConfigAdjustedRule.cs" />
     <Compile Include="Rule\RatNestsSpawnGoldRule.cs" />
     <Compile Include="Rule\SampleRule.cs" />
-    <Compile Include="Rule\PieceConfigAdjustedRule.cs" />
+    <Compile Include="Rule\SorcererStartCardsModifiedRule.cs" />
     <Compile Include="Rule\StartHealthAdjustedRule.cs" />
     <Compile Include="Rule\ZapStartingInventoryAdjustedRule.cs" />
```
orendain, line 53: FYI @jimconner : I'm okay leaving this file indented this way.  My IDE isn't forcing changes, and that way we don't have to worry about fighting VS.

jimconner (commented):
Should we add `using Datakeys;` to RuleMod.cs now that this new ruletype requires it? (Took me a couple of mins to work out why it objected to `Name = AbilityKey.Whirlwind` in the ruleset definition). Maybe that change comes later when we define some actual rulesets... I think we've got enough things working now that we could define a few of them. 

I'm going to hit 'Approve' anyway as I'm sure you'll do the know the right thing to do an will do it before Merging :-)

jimconner (approved):

Rules/Rule/SorcererStartCardsModifiedRule.cs
```diff
@@ -0,0 +1,70 @@
...
+        public struct Card
+        {
+            public AbilityKey Name;
+            public bool IsReplenishable;
+        }
+
+        private static bool _isActivated;
+        private static List<Card> _cards;
+
+        public SorcererStartCardsModifiedRule(List<Card> cards)
```
jimconner, line 23: Could we be passing two arguements to this function, the first one being the name of the Hero to be changing, and the seconf being the list of cards. We don't need to have it all in a single data structure. If we did this then we don't need duplicate classes for adjusting the other heros.

orendain: I agree 100% 👍🏾!  At the moment, each rule type can appear only once in a ruleset (Ruleset stores rules in a `Hashset`), so we'll have to resolve https://github.com/orendain/DemeoMods/issues/83 before we can convert this rule to one that we could use for multiple heroes.

orendain:
Update:  [Implemented the `RulesAPI.IConfigWritable` interface](https://github.com/orendain/DemeoMods/pull/92/commits/c214d97f2589ecc3722bd681f2209d80a12c3e8c), so this rule can now be written to and read from config.

---

> Should we add using Datakeys; to RuleMod.cs now that this new ruletype requires it? (Took me a couple of mins to work out why it objected to Name = AbilityKey.Whirlwind in the ruleset definition). Maybe that change comes later when we define some actual rulesets... I think we've got enough things working now that we could define a few of them.

Yeah, I agree - we can add imports as they're required when rulesets that need it make their way in.

PR #93, opened 2022-02-01
A rule for adjusting the AOE range of abilities

jimconner:
A rule for adjusting the AOE of spells. Works for spells which are already AOE-enabled. Not currently working for others.


orendain (approved):
A bit late on the review, but LGTM!

PR #116, opened 2022-02-06
Pass GameContext in as argument to hook calls.

orendain:
Pass GameContext in as argument to hook calls.

This allows rules that need access to game state to be implementing without capturing game state itself,or patching into parts of the code to change state that it could otherwise do if context was given to it.

This will reduce the likelihood that a rule will need to implement the `IPatchable` interface, in turn becoming simpler to implement.


jimconner (commented):
All makes sense to me.

PR #120, opened 2022-02-06
Add rule modifying card limit.

orendain:
Resolves https://github.com/orendain/DemeoMods/issues/111.

Additionally, moves some logic from the configuration mod to happen during `OnApplicationLateStart()` instead of during `OnApplicationStart()`, so that ruleset loading does not happen before rules are being registered.


jimconner (approved):
Works perfectly. Nice 👍

PR #121, opened 2022-02-06
Fix spelling in demo rule

jimconner:
Fix spelling - because a mis-spelled piece name means that subsequent changes do not get applied.

I had thought that the Assassin was a Rogue-like character, but apparently that is incorrect. The piecename is actually HeroRouge  .. like the costmetics


orendain:
Good catch!

orendain (approved):

PR #127, opened 2022-02-07
Fix card limit rule assuming the card limit was 0 by default.

orendain:
Since the game patch is applied exactly once at game init time, the limit is set to `0` unless the rule is instantiated somewhere in the code before patch time.  Thanks @jimconner for catching this.

This PR increases the max hand limit to a constant (`100` was an arbitrary number), so that the _max_ limit is increased whether or not the rule is used.

That is not to say that the max hand size is always `MaxLimit` - that is just the max limit that the game now allows.  The practical max is still modified in lines `76-81` of the code, and behaves as expect (or at least, it should).


jimconner (approved):
Worked perfectly.

PR #128, opened 2022-02-07
Remove unneeded (?) rules.

orendain:
`ZapStartingInventoryAdjustedRule` was quite a specific rule.  With the addition of `StartCardsModifiedRule`, I wonder if we need to keep `ZapStartingInventoryAdjustedRule` around.

`SorcererStartCardsModifiedRule` may still be useful, if we find that it is too difficult to configure `StartCardsModifiedRule`.

I'm fine either way.

WDYT @jimconner ?


jimconner (approved):
I'd been thinking the same thing myself, and had been planning on getting around to raising exactly this same PR.

PR #131, opened 2022-02-07
Document JSON config options and provide examples for each rule.

jimconner:
Provide a bit more detail about how to use JSON for configuring rules.

I'd like to have included a full example.json file with this, but wasn't sure where the best place for it to reside... Should that be part of Core, Essentials, ConfigMgr or some other dedicated docs/examples location. For now, I've put the file inside Essentials, but not added it to the .csproj... Looking for guidance as to where you'd like to keep this kind of doc.

Whilst working on this I realised that ActionPointsAdjustedRule and StartHealthAdjustedRule are also both covered by PieceConfigAdjustedRule, so potential candidates for deletion.. As such, I didn't document those fully.


HouseRules_Core/README.md
```diff
@@ -29,30 +37,61 @@ If installation instructions were followed (i.e., MelonLoader was installed),
...
 
-A list of out-of-the-box ruleset names can be found in the
-[Rules readme](../RulesAPI_Essentials/README.md).
+Example JSON samples for each rule type can be found in the 
+[Rules readme](../HouseRules_Essentials/README.md).
 
 ## Rules vs Mods
 
-A RulesAPI rule can be written with as few as a couple dozen lines of code,
+A HoiuseRules rule can be written with as few as a couple dozen lines of code,
```
orendain, line 90: nit: small typo.

HouseRules_Essentials/README.md
```diff
@@ -1,35 +1,244 @@
...
+
+### Built-in Rulesets
+
+- **SampleRuleset**: A [sample ruleset](https://github.com/orendain/DemeoMods/blob/045aec568fdddb95b63a1ed34abcb64065e4ca99/Rules/RulesMod.cs#L27-L28)
+  for the purposes of documenting how to create a ruleset (and for testing during development).
+
+### JSON Rulesets
+
+Rulesets may also be defined in JSON within the MelonLoader config file. A sample [MelonLoader config with JSON ruleset](sample-json-ruleset.txt) is provided to help you get started. 
+
```
orendain, line 20: No issue with this text, I just have a related question:

Would it be helpful if we allowed JSON rulesets to exist as their own file?  E.g.: `ruleset = "SampleRuleset"` would imply that the ruleset can be found in a file called `SampleRuleset.json` in some directory (e.g., `/UserData/HouseRules/`)?

Just thinking aloud - figured I'd share before I forgot.

HouseRules_Essentials/README.md
```diff
@@ -1,35 +1,244 @@
...
 ## Rules
 
 - **SampleRule**: A [sample rule](Rules/SampleRule.cs) documenting the anatomy
-  of a RulesAPI rule.
-- **AbilityAOEAdjustedRule**: Ability AOEs are adjsuted
+  of a HouseRule rule.
+- **AbilityAoeAdjustedRule**: Adjusts the Area of Effect range(s) for abilities.
+  - Does not work with all abilities.
+  - Positive integers increase range, negavive decrease. e.g. `"Fireball": 1` will increases the Fireball AOE from a 3x3 to 5x5
```
orendain, line 28: nit: typo in `negavive`.

HouseRules_Core/README.md
```diff
@@ -29,30 +37,61 @@ If installation instructions were followed (i.e., MelonLoader was installed),
...
+
+Rulesets may also be defined in JSON within `UserData/MelonPreferences.cfg`. Here's a ruleset for beefing up the Sorcerer's Zap ability.
+
+```toml
+[HouseRules]
+ruleset = "SuperZapRuleset"
+loadFromConfig = true
+
+[SuperZapRuleset]
+name = "SuperZapRuleset"
```
orendain, line 65: No issue here, just a point for discussion:

It makes sense to have a suffix for `Rule` classes (e.g., `SampleRule.cs`), because of class naming practices.

For rulesets, however, as the names exist really just as a string, I don't think we ought to feel obligated to have the `Ruleset` suffix (outside of class naming).

100% just a side note. 👍🏾

HouseRules_Core/README.md
```diff
@@ -29,30 +37,61 @@ If installation instructions were followed (i.e., MelonLoader was installed),
...
+description = "Zap has no action cost and does 1 extra damage."
+rules = '''
+[
+  {
+    "Rule": "AbilityActionCostAdjustedRule",
```
orendain, line 70: Another tangential thought, not related to the doc:  Now that we've got a good collection of rules, I wonder if we should go back and try to unify and/or improve naming.  I.e.: `AbilityActionCost[[Adjusted]]Rule` -> `Ignored/Waived/Dismissed/Free`.

HouseRules_Essentials/README.md
```diff
@@ -1,35 +1,244 @@
...
+    "Rule": "EnemyDoorOpeningDisabledRule",
+    "Config": true
+  },
+    ```
 - **EnemyHealthScaledRule**: Enemy health is scaled
+    - Config accepts float e.g `0.85`  
+    ###### _Example JSON config for this rule_
+    ```json
+  {
+    "Rule": "EnemyHealthScaledRule",
```
orendain, line 135: Another side note:

While the actual defining class `EnemyHealthScaledRule.cs` has the `Rule` suffix, we could alter the logic of the JSON reader/writer so as to not care about the `Rule` suffix.  Not pushing for one way or another, just mentioning as that thought was sparked.

HouseRules_Essentials/README.md
```diff
@@ -1,35 +1,244 @@
...
+      "FloorThreeHealingFountains": 9,
+      "FloorThreeLootChests": 9
+    },
+    ```
 - **PieceConfigAdjustedRule**: Piece configuration is adjusted
   - See [PieceConfig.md](../docs/PieceConfig.md) for information about modifiable fields.
+  - Allows customization of many of the properties for each game Piece. Health, ActionPoints, Movement, MeleeDamage, etc
+  - Config accepts List of Lists e.g. `[ [ "PieceName1", "ParamName1", int ], [ "PieceName1", "ParamName1", int ], ... ]`   
+  ###### _Example JSON config for this rule_
+  ```json
```
orendain, line 180: Not sure what about this format the GitHub linter doesn't like that it's highlighting it red, but something must be off here.  🤷🏾

orendain (approved):
Woah, nice!  I love it!

Just left some thoughts, mostly tangential/slightly-related as I was reading.  No issues whatsoever from my end with any of these changes - all fantastic 🎉.

HouseRules_Essentials/README.md
jimconner: Oh hell yes... most definitely.. That would be brilliant.... We can put a few examples/built-ins in there... It would be very helpful for new people trying to work out how to create their own rules too.

HouseRules_Essentials/README.md
jimconner: I spotted this when I created the PR and went back and carefully double-checked the file... couldn't see any issue... Checked in VSCode with Markdownlint... It complained about many violations, but nothing specifically around code blocks. I'll fix the markdown warnings tomorrow.

HouseRules_Essentials/README.md
orendain: Implemented here: https://github.com/orendain/DemeoMods/pull/134

HouseRules_Essentials/README.md
jimconner: I agree that having a 'Rule' suffix is pointles here. It would be nicer not to have it.

HouseRules_Essentials/README.md
jimconner: It seems happier about this with the latest commit I just pushed. Was probably an issue with inconsistent indentation. It's much nicer editing markdown in VSCode than in Visual Studio.. I've got all the right plugins in VSCode to highlight my mistakes as I make them.

HouseRules_Essentials/README.md
jimconner: Fixed in latest commit.

HouseRules_Core/README.md
jimconner: This rule can also be used to add an action cost where there currently isn't one. You could make strenght/speed potions cost to cast if desired. Maybe `Toggle/Switch` instead of `Adjusted`, but I don't really see anything wrong with Adjusted in particular instance. I'll keep the thought in mind whilst I'm looking at things though... There may be other opportunites for improving our naming.

HouseRules_Core/README.md
jimconner: I just called the ruleset 'SuperZap' in the latest commit.

orendain (approved):
Incredible work!

PR #132, opened 2022-02-07
Establish how to separate ruleset definitions from one another.

orendain:
**Changes:**
- Add overloaded static method for creating Ruleset without the need to create a list of rules.
  - Instead of having to create a `List<Rule>` before calling `Ruleset.NewInstance(...)`, we can now optionally pass in the rules directly, no matter how many there are.
  - Example:  `Ruleset.NewInstance("MyRuleset", "descriptionHere", myRule1, myRule2, myRule3)`
- Moved sample ruleset to its own file.
  - While rulesets can be defined in very few lines of code, and there's really no functional purpose to moving them to their own file, this makes it easier to find them and maintain them over time.

### DIFF
--- a/HouseRules_Core/Types/Ruleset.cs
+++ b/HouseRules_Core/Types/Ruleset.cs
@@ -1,6 +1,7 @@
 ﻿namespace HouseRules.Types
 {
     using System.Collections.Generic;
+    using System.Linq;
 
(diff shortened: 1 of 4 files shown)


jimconner (approved):
LGTM

PR #134, opened 2022-02-07
Implement full JSON ruleset export/import.

orendain:
As discussed here: https://github.com/orendain/DemeoMods/pull/131#discussion_r801100158

The same flow as before still applies, except:
- `WriteRuleset(myRuleset)` now becomes `ExportRuleset(myRuleset)`.
- `ReadRuleset(rulesetName)` now becomes `ImportRuleset(rulesetName)`.

Stored in `<DemeoDir>/UserData/HouseRules/<ruleset-name>.json`.

### DIFF
--- a/HouseRules_Configuration/ConfigManager.cs
+++ b/HouseRules_Configuration/ConfigManager.cs
@@ -3,7 +3,6 @@
     using System;
     using System.Collections.Generic;
     using System.IO;
-    using System.Security.AccessControl;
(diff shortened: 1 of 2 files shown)


jimconner (approved):
Worked a treat. Wrote config into the right place. Was able to read my own config back in again. Merging this one now, and then I'll update my docs PR to match the new way of handling json.

PR #133, opened 2022-02-07
Add rulesets for easy/hard/legendary difficulty.

orendain:
Resolves https://github.com/orendain/DemeoMods/issues/109.

Marking this PR as a draft until @Buruko gets a chance to try it out and make appropriate tweaks.

Credit to @Buruko for drawing up the following, which is now translated to code:
```
"Difficulty: Easy"
This mode decreases the default game difficulty for a more casual playstyle.

Rules:
EnemyDoorOpeningDisabled
EnemyRespawnDisabled
CardEnergyFromAttackMultiplied (1.5)
EnemyHealthScaled (.8) 20% reduction
EnemyAttackScaled (.6) 40% reduction

"Difficulty: Hard"
This mode increases the default game difficulty for a greater challenge.

Rules:
CardEnergyFromAttackMultiplied (.8)
EnemyHealthScaled (1.2) 20% increase
EnemyAttackScaled (1.2) 20% increase

"Difficutly: Legendary"
This mode increases the default game difficulty for those who want to be a legend.

Rules:
CardEnergyFromAttackMultiplied (.6)
EnemyHealthScaled (1.4) 40% increase
EnemyAttackScaled (1.4) 40% increase
```

PR #146, opened 2022-02-11
Created additional core ruleset

Buruko:
This is an often requested ruleset that is like 'Easy' but without stat modifiers just the specific disabling of doors and respawn.


HouseRules_Essentials/Rulesets/NoSurprisesRuleset.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    using HouseRules.Essentials.Rules;
+    using HouseRules.Types;
+
+    internal static class NoSurprisesRuleset
+    {
+        internal static Ruleset Create()
+        {
+            const string name = "No Surprises.";
+            const string description = "This prevents any surprises in the dark or coming through doors.";
```
orendain, line 11: Small nits:
1. Maybe we shouldn't include trailing periods in the name.  For simplicity/consistency.
2. Instead of `This prevents ...`, WDYT about just `No surp...` or `Prevent any surp...`
  - `This prevents...` just sort of reads like if this were a rule rather than a rulset.

orendain (changes requested):
Cool - nice and simple.  

Check out https://github.com/orendain/DemeoMods/pull/133/files for changes that need to be made to:
- `HouseRules_Essentials/HouseRules_Essentials.csproj` (needed so that the mod recognizes this file as needing to be included in the build)
- ` HouseRules_Essentials/EssentialsMod.cs` (needed so that this ruleset is registered into the rulebook.  Otherwise, this ruleset is just floating around in space)

orendain:
Friendly GitHub tip:  When you want someone to look at your code, request a _review_ from them by adding them as a reviewer in the top-right of GitHub.
image

`Assignees` is typically the person assigned to work on the PR (which is you! 🎉 🎊)

HouseRules_Essentials/Rulesets/NoSurprisesRuleset.cs
Buruko: Agreed and fixed.

Buruko:
My concern was my 'main' fork does not have your edits from your 'eo/difficulty-setting-rules' branch, I figure that GitHub is suppose to sort that out though?

Please pardon my very amateur assumptions and fear of breaking things!

orendain (approved):
> My concern was my 'main' fork does not have your edits from your 'eo/difficulty-setting-rules' branch, I figure that GitHub is suppose to sort that out though?

Yeah, that kind of stuff is handled automatically by git when things are "merged".

If multiple branches make conflicting changes to same file, git will attempt to auto-resolve changes.  If it's not able to, github will complain that the changes you want to make are conflicting, and will require you to resolve those conflicts before the green "Squash and merge" button below even appears.

> Please pardon my very amateur assumptions and fear of breaking things!

No worries!  It's a learning process at first, but fear not - we have some branch protections in place.  Not only must another contributor approve the change (in this case, either me or @jimconner), but even if you manage to merge bad code into the `main` branch, tracking down and undoing things takes seconds/minutes.  👍🏾

> I also believe I will take the rest of the Rulesets I have and create JSON files to be included with the mod as examples.  As they do not fit what I would consider 'core house rules' but ad lib fun stuff to try out.

IMHO, if it's something that you think other people would want to try out, then +1 for having it built-in.  It can always be tweaked/removed in the future if later we decide it no longer fits.

Issue #149, opened 2022-02-12
Determine what all rules are safe (or not) for multiplayer:


orendain:
The rules to either multiplayer label, or not, based on multiplayer testing:

Y = yes
M = no, but might work if board state was resync'd to update values of affected pieces
N = no, host only

AbilityActionCostAdjustedRule.cs - Y, but info isn't updated on card
ActionPointsAdjustedRule.cs - Y
CardEnergyFromAttackMultipliedRule.cs - Y
CardEnergyFromRecyclingMultipliedRule.cs - Y
EnemyDoorOpeningDisabledRule.cs - Y
EnemyRespawnDisabledRule.cs - Y
GoldPickedUpScaledRule.cs - Y
LevelPropertiesModifiedRule.cs - Y
StartCardsModifiedRule.cs - Y

> Note: The following chunk can now be considered safe with the introduction of https://github.com/orendain/DemeoMods/pull/151.

AbilityAOEAdjustedRule.cs - M
AbilityDamageAdjustedRule.cs - M
EnemyAttackScaledRule.cs - M
EnemyHealthScaledRule.cs -M
PieceConfigAdjustedRule.cs - M
StartHealthAdjustedRule.cs   - M

CardLimitModifiedRule.cs - N
CardSellValueMultipliedRule.cs - N
RatNestsSpawnGoldRule.cs - N

orendain:
CC @jimconner @Buruko

orendain:
Closing this as it seems to have been resolved.  Feel free to open another issue if piece stats aren't syncing up as expected, or something non-multiplayer safe needs addressing.

PR #151, opened 2022-02-12
Force board state resync in cases where a client may register a new spawn.

orendain:
Related to https://github.com/orendain/DemeoMods/issues/149.

Force board state resync in cases where a client may register a new spawn.

Existing rules that affect the stats of a piece are mostly resolved client-side. For the host's overridden piece stats to propagate to clients, the host needs to somehow sync its state with each client.

It is important to trigger this sync as close to the spawn event as possible, else the client may interact with the piece given non-actual stats, potentially causing desync issues.

Due to how heavy the resync is, it is undesirable to naively sync periodically (e.g., at the end of each action). However, because Demeo has only fine-grained and super-heavy-grained sync events, some overhead is to be expected.

### DIFF
--- a/HouseRules_Core/HR.cs
+++ b/HouseRules_Core/HR.cs
@@ -11,15 +11,16 @@ namespace HouseRules
     {
         internal static readonly MelonLogger.Instance Logger = new MelonLogger.Instance("HouseRules:Core");
         private const float WelcomeMessageDurationSeconds = 30f;
-        private static bool _isRulesetActive;
 
         public static readonly Rulebook Rulebook = Rulebook.NewInstance();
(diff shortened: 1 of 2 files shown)


orendain:
I had to perform quite a few granular checks so as to be careful not to force-sync in cases where it wouldn't be required.  There wasn't much middle ground that I could find ... it was either we force-sync too frequently (which means way too much load on the network), or we force-sync less frequently but it leaves desync opportunities open to clients (because they may go a round or two with the wrong board and piece information).

So I spent some extra time curating a special hook that will trigger a force-sync only when a spawn may have been possible.

PR #160, opened 2022-02-13
Updated Difficulty Rulesets and added ruleset

Buruko:
Updated the Difficulty rulesets and added a new ruleset with a description as Skirmish only.

We will need to label any ruleset that contains on of the three as Skirmish unless we can toss out rulesets on load if they are Multiplayer. We don't want the mod to break someone's game cause they don't understand the Ruleset and then blame the mod.


HouseRules_Essentials/Rulesets/QuickandDeadRuleset.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            const string name = "Quick and the Dead";
+            const string description = "A Skirmish mode with a small hand but fast turn around time on cards means you need to not hesitate.";
+
+            return Ruleset.NewInstance(
+                name,
+                description,
+                new CardEnergyFromAttackMultipliedRule(1.5f),
+                new CardEnergyFromRecyclingMultipliedRule(1.5f),
+                new CardLimitModifiedRule(7);
```
orendain, line 18: I believe this is missing an extra `)` right before the semicolon.

orendain (changes requested):
Could we also add this file to `HouseRules_Essentials/HouseRules_Essentials.csproj`?  For example, somewhere around line `41` in that file.  So that it gets compiled ("bundled in with") the mod when the the mod gets built into a DLL.

Regarding labeling as multiplayer:
- Individual rules must now be "labeled" as being multiplayer safe (they default to not multiplayer-safe if they are not labeled).  For example, some labelling happened here:
  - https://github.com/orendain/DemeoMods/pull/150
  - https://github.com/orendain/DemeoMods/pull/152
- When a multiplayer game is started, HouseRules will scan the selected ruleset.  If the ruleset contains any rules that are not marked safe for multiplayer, it will not load the ruleset.  Additionally, it will `Display a message on game load when a ruleset was not activated due to being unsafe for multiplayer environments.`.  That check was introduced in the following PR: https://github.com/orendain/DemeoMods/pull/147

Instead of making it the responsibility of _each ruleset description_ to let players know if it's multiplayer-safe or not, I think we ought to tweak the UI so that somehow we get the point across.  Maybe (random idea, for example, not saying it's a good one) we split rulesets into 2 sections (multiplayer-safe and not) or have a checkmark next to each ruleset denoting if its safe.

That way, ruleset descriptions can be just ruleset descriptions (and we make it shorter, because the current UI is very narrow 😅).

HouseRules_Essentials/Rulesets/QuickandDeadRuleset.cs
Buruko: Yes! Corrected.

Perhaps sections for Skirmish and Multiplayer rules in a Stack lay out would be better but then wouldn't we need a way to tag a ruleset to fall under each 'group' ?

orendain: When a ruleset is created, all of its rules are checked.  If they're all multiplyaer safe, then the ruleset is multiplayer safe as well.  If not, then the ruleset is marked as not safe for multiplayer.  https://github.com/orendain/DemeoMods/blob/main/HouseRules_Core/Types/Ruleset.cs#L49

So there is indeed now a way to check if a ruleset is safe! 👍🏾

orendain (approved):
LGTM!

PR #164, opened 2022-02-16
Add rule to override StatusEffectConfig 

jimconner:
# New Rule to override StatusEffectConfig

The behaviour of the status effects such as Torch, Poisoned, Weakened, SongOfHealing, etc are configured through a list of StatusEffectData. This PR implements a new rule which accepts a similar list of overriding StatusEffectData and patches the `GetConfigFor` function so that it will return any matching overrides if present. If not then default config will be used.

Please note that this has not been testing in multiplayer yet.

## Suggested JSON for testing.
```
{
  "Name": "LuckyDip",
  "Description": "Life is like a box of chocolates + AOE changes, so stay close to your allies",
  "Rules": [
      {
        "Rule": "AbilityDamageAdjusted",
        "Config": { "Zap": 1 }
      },
      {
        "Rule": "StartCardsModified",
        "Config": {
          "HeroSorcerer": [
            { "Card": "Heal", "IsReplenishable": false },
            { "Card": "Zap", "IsReplenishable": true },
            { "Card": "TorchLight", "IsReplenishable": true },
            { "Card": "DropChest", "IsReplenishable": false },
            { "Card": "DropChest", "IsReplenishable": false }
          ],
          "HeroBard": [
            { "Card": "Heal", "IsReplenishable": false },
            { "Card": "StrengthenCourage", "IsReplenishable": true },
            { "Card": "SongOfRecovery", "IsReplenishable": true },
            { "Card": "DropChest", "IsReplenishable": false },          
            { "Card": "DropChest", "IsReplenishable": false },
            { "Card": "ShatteringVoice", "IsReplenishable": false }
          ]
        }
      },
    {
      "Rule": "StatusEffectConfig",
      "Config": [
        {
          "effectStateType": "TorchPlayer",
          "durationTurns": 15,
          "tickWhen": "StartTurn",
          "stacks": true,
          "damagePerTurn": 0,
          "clearOnNewLevel": false,
          "damageTags": null,
          "healPerTurn": 0
        },
        {
          "effectStateType": "HealingSong",
          "durationTurns": 4,
          "tickWhen": "StartTurn",
          "stacks": false,
          "damagePerTurn": 0,
          "clearOnNewLevel": false,
          "damageTags": null,
          "healPerTurn": 3
        }
      ]
    }
  ]
}
```

### DIFF
--- a/HouseRules_Essentials/Rules/StatusEffectConfigRule.cs
+++ b/HouseRules_Essentials/Rules/StatusEffectConfigRule.cs
@@ -1,0 +1,69 @@
+﻿namespace HouseRules.Essentials.Rules
+{
+    using System;
+    using Boardgame;
+    using DataKeys;
+    using HarmonyLib;
+    using HouseRules.Types;
+
+    public sealed class StatusEffectConfigRule : Rule, IConfigWritable<global::Types.StatusEffectData[]>, IPatchable, IMultiplayerSafe
+    {
+        public override string Description => "Status Effect Durations are modified";
+
+        private static bool _isActivated;
+        private static global::Types.StatusEffectData[] _adjustments;
+
+        public StatusEffectConfigRule(global::Types.StatusEffectData[] adjustments)
+        {
+            _adjustments = adjustments;
+        }
+
+        public global::Types.StatusEffectData[] GetConfigObject() => _adjustments;
+
+        protected override void OnActivate(GameContext gameContext)
+        {
+            _isActivated = true;
+
+        }
+
+        protected override void OnDeactivate(GameContext gameContext) => _isActivated = false;
+
+        private static void Patch(Harmony harmony)
+        {
+            harmony.Patch(
+                original: AccessTools.Method(typeof(StatusEffectsConfig), "GetConfigFor"),
+                prefix: new HarmonyMethod(
+                    typeof(StatusEffectConfigRule),
+                    nameof(StatusEffects_GetConfigFor_Prefix)));
+        }
+
+        private static bool StatusEffects_GetConfigFor_Prefix(ref EffectStateType type, ref global::Types.StatusEffectData __result)
+        {
+            if (!_isActivated)
+            {
+                return true;
+            }
+
+            var effectsConfig = Traverse.Create(typeof(StatusEffectsConfig)).Field<global::Types.StatusEffectData[]>("effectsConfig").Value;
+
+            for (int i = 0; i < _adjustments.Length; i++)
+            {
+                if (_adjustments[i].effectStateType == type)
+                {
+                    __result = _adjustments[i];
+                    return false; // We returned an user-adjusted config.
+                }
+            }
+            for (int i = 0; i < effectsConfig.Length; i++)
+            {
+                if (effectsConfig[i].effectStateType == type)
+                {
+                    __result = effectsConfig[i];
+                    return false; // We returned original result.
+                }
+            }
+
+            throw new Exception("Could not find config for " + type);
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


HouseRules_Essentials/Rules/StatusEffectConfigRule.cs
```diff
@@ -0,0 +1,69 @@
...
+    using HouseRules.Types;
+
+    public sealed class StatusEffectConfigRule : Rule, IConfigWritable<global::Types.StatusEffectData[]>, IPatchable, IMultiplayerSafe
+    {
+        public override string Description => "Status Effect Durations are modified";
+
+        private static bool _isActivated;
+        private static global::Types.StatusEffectData[] _adjustments;
+
+        public StatusEffectConfigRule(global::Types.StatusEffectData[] adjustments)
```
orendain, line 16: Generally, lists (e.g. `List<StatusEffectData>`) are preferred over arrays (See [this item from Effective Java](https://www.informit.com/articles/article.aspx?p=2861454&seqNum=3), although the same (mostly) applies to CSharp).

Lists will also un/marshal into JSON the same as arrays.

HouseRules_Essentials/Rules/StatusEffectConfigRule.cs
```diff
@@ -0,0 +1,69 @@
...
+            for (int i = 0; i < effectsConfig.Length; i++)
+            {
+                if (effectsConfig[i].effectStateType == type)
+                {
+                    __result = effectsConfig[i];
+                    return false; // We returned original result.
+                }
+            }
+
+            throw new Exception("Could not find config for " + type);
```
orendain, line 66: Since the original code already performs this check, the patch could instead `return true` here (which would only happen if we haven't already returned `false` by overriding the result ourselves).

Returning true would allow the original code to run (after ours does), where it would do this for us.  The benefit would be that we don't need to update this code when/if the original code changes in the future.

orendain (approved):
LGTM!  This now enables #5.

---

Downsides of patching:
- Due to patching limitations (patched methods must be `static`) and how core is designed, rules with patches can exist only once in a given ruleset.
- Our custom functions are called each time the original is, adding (in most cases negligible) overhead.

Since `getConfigFor()` performs a lookup of `effectsConfig`, we could instead modify `effectsConfig` directly during say, `OnActivate()`.  Since we modify the collection itself, the changes would persist until we undo it.

Additionally, since no patching would be required, the rule could be used multiple times in a single ruleset (probably not much of a benefit, but figured I'd mention it).

---

All of my comments are really just nits or info for knowledge share.  This is great as is.  Will merge in later today (in case any changes are made - but not necessary).

HouseRules_Essentials/Rules/StatusEffectConfigRule.cs
```diff
@@ -0,0 +1,69 @@
...
+    using DataKeys;
+    using HarmonyLib;
+    using HouseRules.Types;
+
+    public sealed class StatusEffectConfigRule : Rule, IConfigWritable<global::Types.StatusEffectData[]>, IPatchable, IMultiplayerSafe
+    {
+        public override string Description => "Status Effect Durations are modified";
+
+        private static bool _isActivated;
+        private static global::Types.StatusEffectData[] _adjustments;
```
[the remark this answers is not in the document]
jimconner: Will fix in next commit.

HouseRules_Essentials/Rules/StatusEffectConfigRule.cs
jimconner: Resolved in https://github.com/orendain/DemeoMods/pull/164/commits/bdfa98024eeffde696e67bff9cd2ac0250ee2bfb

HouseRules_Essentials/Rules/StatusEffectConfigRule.cs
jimconner: Resolved in https://github.com/orendain/DemeoMods/pull/164/commits/bdfa98024eeffde696e67bff9cd2ac0250ee2bfb

PR #171, opened 2022-02-17
Tweak BeatTheClock ruleset.

orendain:
- Reduced HP to 200 to make health depletion at least somewhat more interesting.
- Reduced round count to 15 to increase difficulty.
- Decreased card recycling to reduce spam.
- Added POIs to increase difficulty by way of hiding the exit.


jimconner (approved):
15 Rounds is going to be a LOT more challenging. Looking forward to trying this out.

PR #172, opened 2022-02-17
Use item's key when intepolating strings.

orendain:
PR was originally for:
2b577e5 Use item's key when intepolating strings.

But in the process, I figured I'd boyscout:
fddec51 Use BoardPieceId type instead of string for static type checking.

Using the static type `BoardPieceId` in place of a string allows us better compile-time checking, and will fail-fast when reading in a ruleset from JSON, where the string the user provided does not map to a boardpieceId (instead of allowing the rule to continue, and then only failing during rule activation).


jimconner:
I'll make similar changes to PieceConfigAdustedRule soon too.

jimconner (approved):
LGTM

PR #173, opened 2022-02-17
PieceConfigAdjustedRule config uses desired values instead of +- modifiers.

jimconner:
# What
This PR switches the PieceConfigAdjustedRule from using +- modifers to the base values and instead replaces the defaults with the settings.

It was confusing making rulesets with the +- modifiers. If you decided that you wanted the CaveTroll to only have 5HP, you'd first have to know that it has 16 by default and then use a -11 modifier. It's far more straightforward just to specify the desired result.


orendain (approved):
Fantastic change!

PR #181, opened 2022-02-17
Import all JSON rulesets on game load.

orendain:


jimconner (approved):
Works like a charm. Perfect.

PR #187, opened 2022-02-18
Add PetsFocusHunterMarkRule.

orendain:
Pets focus on hunter marked enemies.

Config:
```json
{
    "Rule": "PetsFocusHunterMark",
    "Config": true,
},
```


jimconner:
What a great rule.. I like it a lot :-)  

Does need docs though.

PR #210, opened 2022-02-22
Two of the Boss characters are inconsistently named.

jimconner:
Two of the pieces have PieceConfig names which differ from their BoardPieceID names.

DarkElfGoddessBoss is the BoardPieceID, but DarkElfGoddess_Boss is the pieceConfig name.
Similarly with CaveTrollBoss and CaveTroll_Boss

I tried switching things over to using the `PieceConfigDTOdict` instead, but either making modifications to that structure didn't have the desired effect, or I was doing it wrong.

This was the least-bad way I could find for fixing the boss names.

### DIFF
--- a/HouseRules_Essentials/Rules/PieceAbilityListOverriddenRule.cs
+++ b/HouseRules_Essentials/Rules/PieceAbilityListOverriddenRule.cs
@@ -31,9 +31,10 @@
         protected override void OnPostGameCreated(GameContext gameContext)
         {
             var pieceConfigs = Resources.FindObjectsOfTypeAll<PieceConfig>();
+            string lookupstring = string.Empty;
             foreach (var item in _adjustments)
             {
-                var pieceConfig = pieceConfigs.First(c => c.name.Equals($"PieceConfig_{item.Key}"));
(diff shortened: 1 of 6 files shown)


HouseRules_Essentials/Rules/PieceAbilityListOverriddenRule.cs
```diff
@@ -31,9 +31,10 @@ public PieceAbilityListOverriddenRule(Dictionary<BoardPieceId, List<AbilityKey>>
         protected override void OnPostGameCreated(GameContext gameContext)
         {
             var pieceConfigs = Resources.FindObjectsOfTypeAll<PieceConfig>();
+            string lookupstring = string.Empty;
```
orendain, line 34: `lookupstring` looks unused in all of the files.

orendain (approved):
LGTM other than unused `lookupstring` variable in the files.

HouseRules_Essentials/Rules/PieceAbilityListOverriddenRule.cs
jimconner: Good spot - that was a left-over from before I moved the FixBossNames method to a common location.

PR #211, opened 2022-02-22
SpawnCategoryOverridden,  CardClassRestrictionOverridden & Arachnophobia Ruleset.

jimconner:
# A new Arachnophobia Ruleset, and two new rules required to support it.

- __SpawnCategoryOverridden__ - Replaces the list of SpawnCategories from which the MonsterDecks for each level are constructed. This makes it possible to control which types of monsters are allowed in which 🗺️levels.
- __CardClassRestrictionOverridden__ - Replaces the ClassRestriction property for any given 🃏card(s). This can be used to assign 🃏cards to different players, or disable them from assignment by changing the class restriction to a non-player item (e.g. mushroom).

## 🕷️Arachnophobia🕷️ Ruleset
_💰Money 🕷️spiders everywhere, on the walls and in my hair_ - This ruleset is intended for play on the RootsOfEvil 🗺️map. The Ruleset replaces the default enemies, changes some enemy abilities and behaviours and gives the ⚔️heros some new powers to help them in their struggle against the baddies.

Please note that this PR includes the same 'Two of the Boss characters are inconsistently named` commit that I PR'd earlier. Don't know if this will cause any kind of merge conflict.. I needed the commit in this branch to test my ruleset with.

## How to test.
Use the included `Arachnophobia.json` ruleset

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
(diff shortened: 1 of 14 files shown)


orendain (approved):
LGTM!  Excited to try this out at some point.

Will leave merging to you because of the potential merge conflict.

PR #212, opened 2022-02-23
Converted to Collapsible JSON examples

Buruko:
Fecn mentioned collapsible JSON examples, I found a way to do this view the Wiki in markdown so have updated the Essentials ReadME to have collapsing examples.

### DIFF
--- a/HouseRules_Essentials/README.md
+++ b/HouseRules_Essentials/README.md
@@ -33,8 +33,9 @@ The [Settings Reference](../docs/SettingsReference.md) contains lists of all dif
   - Overrides the Ability.CostAP setting for player abilities.
   - `true` means the ability has a cost to cast, `false` means that it doesn't.
   - Config accepts Dictionary e.g. `{ "AbilityName1": bool, "AbilityName2": bool, }`
-  
-  ###### _Example JSON config for AbilityActionCostAdjustedRule_
+
+  <details>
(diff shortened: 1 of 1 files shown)


Buruko:
You can see a working example at the end of the HouseRules Wiki page: https://github.com/orendain/DemeoMods/wiki/House-Rules

HouseRules_Essentials/README.md
```diff
@@ -263,26 +293,30 @@ The [Settings Reference](../docs/SettingsReference.md) contains lists of all dif
     }
   },
   ```
+  </default>
```
jimconner, line 296: Shouldn't this be a </default> tag? Currently wen you click the 'show json' tag for this, it mashes up the results.

jimconner:
This approach with the details tags was one that I tried and decided not to use -  the markdown linter in VSCode didn't like the HTML tags in Markdown, and that made it hard to spot the real markdown errors... as such, I decided to go with just plain markdown, and found a format that I thought looked good for the documentation.  As this PR re-implements a style I moved away from, I figure the final decision should go to someone else. I created the docs in the kind of format that I like to consume them from, but ultimately it shouldn't be about how I want the docs to look.. it should be about how users find them to be most easy to consume.  Removing myself as a reviewer.

jimconner:
I'm going to close this PR down because the docs have moved on since then. Thank you very much for your efforts @Buruko - sorry this PR didn't make it through.

Issue #213, opened 2022-02-24
In-game menus for RoomFinder and Houserules are not displayed after exiting Heros Hangout


jimconner:
V 1.12.143299 of Demeo released a few minutes ago, so I gave it a quick test to check.

I have discovered that when exiting a private Heros Hangout session you are returned to the main menu but it no longer displays the RoomFinder or HouseRules menus.

PR #214, opened 2022-02-25
Allow rediscovering Demeo resources, in case they are recreated.

orendain:
Resolves https://github.com/orendain/DemeoMods/issues/213.

The issue was that after leaving hangouts, Demeo unloads the lobby scene.  That isn't the case when leaving a game (e.g., completing a game takes you _back_ to the lobby, which was never unloaded).

As a result of the lobby being unloaded, some assets that the Common/UI module expected to be static were not.  And as a result of _that_, both HouseRules and RoomFinder UIs were not able to depend on the location of a lobby-positioned Demeo resource to latch onto for display in the main lobby.

These Demeo assets must be dynamically capturable instead of assuming to be static (because now we know the scene may indeed be unloaded).

This PR makes that so, and also adds an `EnsureResourcesExists()` check that will recapture the value of required Demeo assets when one of the mods uses the Common/UI helper (which fixes HouseRules and RoomFinder).

### DIFF
--- a/Common/UI/DemeoResource.cs
+++ b/Common/UI/DemeoResource.cs
@@ -13,30 +13,21 @@
 
         public Color ColorBeige { get; } = new Color(0.878f, 0.752f, 0.384f, 1);
 
-        public Component LobbyAnchor { get; } = Resources.FindObjectsOfTypeAll<charactersoundlistener>()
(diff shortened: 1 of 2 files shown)


jimconner (approved):
Works perfectly.

PR #215, opened 2022-02-25
Cleanup - remove unused 'lookupstring' var

jimconner:
Variable was added as part of work to add underscores into boss names before the method got moved to a central location. These vars were missed in the cleanup.


orendain (approved):
LGTM.

PR #216, opened 2022-02-26
Control whether cards are returned to players hand when cast on someone already at max.

jimconner:
A new rule to control whether strenght/speed ability cards are returned to the player's hand when they are cast on something that is already at Max.

I gave this a quick test by assigning the sorcerer 4x speed cards in his starting hand. All appears to work as expected.


orendain (approved):
LGTM

PR #217, opened 2022-02-27
Docs did not show that CardLimit rule was skirmish only.

jimconner:
This caused some confusion for a player as they were trying play multiplayer with this rule.


orendain (approved):
LGTM

PR #218, opened 2022-03-06
Improved rule for StatusEffectsConfig.

jimconner:
The use of Harmony.Patch with this rule was undesireable. Having multiple rulesets loaded which all used StatuesEffectsConfig would cause one ruleset to overwrite values from others.

This version of the rule accesses the StatusEffectsData[] array directly instead.


orendain (approved):
Great enhancement!

PR #219, opened 2022-03-07
Allow backstab to be configured for multiple pieces.

jimconner:
The backstab bonus is applied based on whether the `isRogue()` function within a `Piece` returns true or false. `IsRogue()` doesn't get used anywhere else apart from Backstab bonus, so shouldn't have any unexpected consequences.

Test JSON:
```
    {
      "Rule": "BackstabConfigOverridden",
      "Config": [ "HeroGuardian", "HeroHunter", "HeroSorcerer", "HeroRouge", "HeroBard"  ]
    },
```

I'm guessing that there is probably a better way of doign this so that rules don't overwrite each other's settings.... Possibly with the patch to make isRogue take a list of IDs being separate from the rule that configures the list... but that's beyond my ability to implement at the moment.

### DIFF
--- a/HouseRules_Essentials/Rules/BackstabConfigOverriddenRule.cs
+++ b/HouseRules_Essentials/Rules/BackstabConfigOverriddenRule.cs
@@ -1,0 +1,62 @@
+﻿namespace HouseRules.Essentials.Rules
+{
+    using System.Collections.Generic;
+    using Boardgame;
+    using Boardgame.BoardEntities;
+    using DataKeys;
+    using HarmonyLib;
+    using HouseRules.Types;
+    using StatusEffectData = global::Types.StatusEffectData;
+
+    public sealed class BackstabConfigOverriddenRule : Rule, IConfigWritable<List<BoardPieceId>>, IPatchable, IMultiplayerSafe
+    {
+        public override string Description => "Backstab is allocated to new BoardPieceIDs";
+
+        private static List<BoardPieceId> _globalAdjustments;
+        private static bool _isActivated;
+        
+        private static List<BoardPieceId> _adjustments;
+
+        public BackstabConfigOverriddenRule(List<BoardPieceId> adjustments)
+        {
+            _adjustments = adjustments;
+        }
+
+        public List<BoardPieceId> GetConfigObject() => _adjustments;
+
+        protected override void OnActivate(GameContext gameContext)
+        {
+            _globalAdjustments = _adjustments;
+            _isActivated = true;
+        }
+
+        protected override void OnDeactivate(GameContext gameContext) => _isActivated = false;
+
+        private static void Patch(Harmony harmony)
+        {
+            harmony.Patch(
+                original: AccessTools.Method(typeof(Piece), "IsRogue"),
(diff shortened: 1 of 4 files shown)


jimconner:
Maybe this needs a different rule name... There is also the setting for whether an ability is capable of backstab. Perhpas this should be called PieceBackstabConfigRule instead?

HouseRules_Essentials/Rules/BackstabConfigOverriddenRule.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        private static bool Piece_IsRogue_Prefix(ref Piece __instance, ref bool __result)
+        {
+            if (!_isActivated)
+            {
+                return true;
+            }
+
+            if (_adjustments.Contains(__instance.boardPieceId))
```
orendain, line 48: > I'm guessing that there is probably a better way of doign this so that rules don't overwrite each other's settings.... Possibly with the patch to make isRogue take a list of IDs being separate from the rule that configures the list... but that's beyond my ability to implement at the moment.

Does the following info help?

---

We may want to introduce a static/global field (e.g., like `_globalMultiplier` in `CardEnergyFromRecyclingMultipliedRule`), instead of using `_adjustments`, so that multiple versions of this rule don't interfere with each other.  I think all other `IPatchable` rules already use some `_global` field.

Example issue:
- Ruleset created and uses `new BackstabConfigOverriddenRule(myFieldsOne);`.  That sets the static `_adjustments` to `myFieldsOne`.
- Another ruleset created and uses `new BackstabConfigOverriddenRule(myFieldsTwo);`.  That sets (**_overrides_**) the static `_adjustments` to `myFieldsTwo`
- Even if the first ruleset is loaded, the latest instantiation of the rule (using `myFieldsTwo`) will apply.

Using a static `_global` field is an unfortunate (but simple/elegant) workaround and originates from a limitation in Harmony that a patch method must be static.  Static patch method implies using static fields.

orendain (approved):
Aside from a small issue with using a global/static field, this LGTM for what it's trying to do (makes all melee attacks from all heroes backstab-able).

> Maybe this needs a different rule name... There is also the setting for whether an ability is capable of backstab. Perhpas this should be called PieceBackstabConfigRule instead?

Seems like a fair idea for enabling backstab damage for any attack.

IMHO, I still like this rule as is.  Seems like a simple, straightforward way to enable melee backstabs for all heroes without having to go into detail on specific abilities.

HouseRules_Essentials/Rules/BackstabConfigOverriddenRule.cs
jimconner: Thank you - that was exactly what I was looking for. I've now added a similar _globalAdjustments and given it a test. All appears to work correctly.

jimconner: I decided that the rulename was OK for this. It is not adjusting a PieceConfig property, and doesn't break our naming convention for ability rules. I'm gonna merge now.

PR #226, opened 2022-03-09
Rule to control the enableBackstabBonus toggle on Abilities.

jimconner:
In #219  I'd pondered whether the rule name chosen was suitable or whether it would clash with a rule for changing the backstab setting on abilities. At that time, I hadn't actually been successful in changing an ability's backstab through UnityExplorer. As there was already a near perfect rule to copy, I quickly put together a new one for adjusting an ability's backstab bonus setting.

During the process of putting this rule together, I realised that there would be no naming clash between different backstab rules, so merged the other PR.

Having put a new rule together, I figured I might as well PR it. 

This rule is very lazy... There are so many other bools that we could toggle, it is probably worthwhile putting together a more generic rule along the lines of PieceConfigAdjusted which can accept param names and values instad.

## Test JSON
```
    {
      "Rule": "AbilityBackstabAdjustedRule",
      "Config": {
        "Zap": true,
        "HunterArrow": true,
        "PiercingArrow": true,
        "PoisonTip": true,
        "Fireball": true,
        "Freeze": true,
      }
    },
```

### DIFF
--- a/HouseRules_Essentials/Rules/AbilityBackstabAdjustedRule.cs
+++ b/HouseRules_Essentials/Rules/AbilityBackstabAdjustedRule.cs
@@ -7,18 +7,18 @@
     using HouseRules.Types;
     using UnityEngine;
 
-    public sealed class AbilityActionCostAdjustedRule : Rule, IConfigWritable<Dictionary<string, bool>>, IMultiplayerSafe
+    public sealed class AbilityBackstabAdjustedRule : Rule, IConfigWritable<Dictionary<string, bool>>, IMultiplayerSafe
     {
         public override string Description => "Ability AP costs are adjusted";
 
         private readonly Dictionary<string, bool> _adjustments;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="AbilityActionCostAdjustedRule"/> class.
+        /// Initializes a new instance of the <see cref="AbilityBackstabAdjustedRule"/> class.
         /// </summary>
-        /// <param name="adjustments">Key-value pairs of abilitykey and bool that controls 
-        /// the costActionPoint setting.</param>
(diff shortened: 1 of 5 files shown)


HouseRules_Essentials/Rules/AbilityBackstabAdjustedRule.cs
```diff
@@ -0,0 +1,48 @@
...
+    using Boardgame;
+    using Boardgame.BoardEntities.Abilities;
+    using HouseRules.Types;
+    using UnityEngine;
+
+    public sealed class AbilityBackstabAdjustedRule : Rule, IConfigWritable<Dictionary<string, bool>>, IMultiplayerSafe
+    {
+        public override string Description => "Ability AP costs are adjusted";
+
+        private readonly Dictionary<string, bool> _adjustments;
```
orendain, line 14: QQ: Is there a reason for accepting `string` instead of `AbilityKey`?  I feel like there was talk about this in the past but I can't recall.

HouseRules_Essentials/Rules/AbilityBackstabAdjustedRule.cs
```diff
@@ -0,0 +1,48 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using Boardgame;
+    using Boardgame.BoardEntities.Abilities;
+    using HouseRules.Types;
+    using UnityEngine;
+
+    public sealed class AbilityBackstabAdjustedRule : Rule, IConfigWritable<Dictionary<string, bool>>, IMultiplayerSafe
+    {
+        public override string Description => "Ability AP costs are adjusted";
```
orendain, line 12: nit:  Different description.

orendain (approved):
LGTM, only minor things to point out.

Will make some minor changes tonight - will include a couple of the fixes below in it.

Issue #220, opened 2022-03-09
Provide some visual feedback when selecting a ruleset ingame.


orendain:
Currently, we write to logs in addition to the stock audio feedback when selecting UI buttons.  However, there is no visual indication that a ruleset was selected via the UI.

PR #228, opened 2022-03-10
Provide visual feedback of selected ruleset on UI.

orendain:
Resolves https://github.com/orendain/DemeoMods/issues/220.

**Changes:**

- Display selected ruleset on UI.
- Reduce gap between listed rulesets.
- Rename fields and labels slightly.
- Rename widgets.

**Notes:**

Originally, I wanted to save some space and give the UI that extra bit of pizzazz by having buttons alternate colors when selected.  While solved, implementation would have taken significantly longer than this approach of simply displaying the selected ruleset.  The original idea can still be implemented, but for the sake of saving some time this was the approach this time around.

**Screenshot to demonstrate new UI:**

image

### DIFF
--- a/Common/UI/UiHelper.cs
+++ b/Common/UI/UiHelper.cs
@@ -11,7 +11,8 @@
         public const int DefaultButtonFontSize = 7;
         public const int DefaultLabelFontSize = 5;
         public const int DefaultMenuHeaderFontSize = 10;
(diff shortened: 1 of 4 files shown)


jimconner (approved):
All looks good. Works as advertised. Nice work Mex 👍

PR #229, opened 2022-03-10
Adjust RoomFinder UI to compensate for common UI changes.

orendain:
Adjust RoomFinder UI to compensate for common UI changes.

Should ideally not look different than before the day's changes.

> Note:  Should be merged after https://github.com/orendain/DemeoMods/pull/228.


jimconner (approved):
I can't see any visible differences, which I guess means that this was a success :-)

PR #231, opened 2022-03-10
Difficulty Adjustments

Buruko:
Changed scaling for HP/ATK on higher Difficulty

Added `LevelPropertiesModified` rule to reduce resources at higher difficulties

### DIFF
--- a/HouseRules_Essentials/Rulesets/DifficultyHardRuleset.cs
+++ b/HouseRules_Essentials/Rulesets/DifficultyHardRuleset.cs
@@ -10,14 +10,30 @@
             const string name = "Difficulty: Hard";
             const string description = "Increased game difficulty for a greater challenge.";
 
+            var cardEnergyAttck = new CardEnergyFromAttackMultipliedRule(0.8f);
+            var cardEnegyRecycle = new CardEnergyFromRecyclingMultipliedRule(0.8f);
+            var EnemyScaleHealth = new EnemyHealthScaledRule(1.8f);
+            var EnemyScaleAttack = new EnemyAttackScaledRule(1.5f);
+
+            var levelPropertiesModified = new LevelPropertiesModifiedRule(new Dictionary<string, int>
+            {
+              { "BigGoldPileChance", 30 },
+              { "FloorOneHealingFountains", 1 },
+              { "FloorOneLootChests", 2 },
+              { "FloorTwoHealingFountains", 1 },
+              { "FloorTwoLootChests", 2 },
+              { "FloorThreeHealingFountains", 1 },
+              { "FloorThreeLootChests", 1 },
+            });
+
             return Ruleset.NewInstance(
                 name,
                 description,
-                new CardEnergyFromAttackMultipliedRule(0.8f),
-                new CardEnergyFromRecyclingMultipliedRule(0.8f),
-                new EnemyHealthScaledRule(1.8f),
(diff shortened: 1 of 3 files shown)


HouseRules_Essentials/Rulesets/DifficultyHardRuleset.cs
```diff
@@ -15,8 +15,9 @@ internal static Ruleset Create()
                 description,
                 new CardEnergyFromAttackMultipliedRule(0.8f),
                 new CardEnergyFromRecyclingMultipliedRule(0.8f),
-                new EnemyHealthScaledRule(1.2f),
-                new EnemyAttackScaledRule(1.2f));
+                new EnemyHealthScaledRule(1.8f),
+                new EnemyAttackScaledRule(1.5f),
+                new LevelPropertiesModifiedRule("BigGoldPileChance", 30, "FloorOneHealingFountains", 1, "FloorOneLootChests", 2, "FloorTwoHealingFountains", 1, "FloorTwoLootChests", 2, "FloorThreeHealingFountains", 1, "FloorThreeLootChests", 1));
```
orendain, line 20: Values are passed into the `LevelPropertiesModifiedRule` rule in a _slightly_ different way.  Here's an example from The Swirl ruleset.  Note that values are defined more like pairs than all in a single line separated by commas:

https://github.com/orendain/DemeoMods/blob/118ad9a30668ce51d8211bfb89f520aa8377b4df/HouseRules_Essentials/Rulesets/TheSwirlRuleset.cs#L68-L76

orendain (commented):
Specifying values for `LevelPropertiesModifiedRule` rule needs a tiny bit of tweaking.

Guessing the readme changes were leftover from another PR.

HouseRules_Essentials/Rulesets/DifficultyHardRuleset.cs
Buruko: Ah. I see whatcha mean. Will correct and re-submit.

HouseRules_Essentials/README.md
```diff
@@ -507,7 +554,8 @@ The [Settings Reference](../docs/SettingsReference.md) contains lists of all dif
   - Max of two replenishable cards per player.
   - Config accepts Dictionary of list of dicts e.g. `{ "HeroName1": [ { "Card" : "CardName","isReplenishable": bool }, ... ], ...  }`
 
-  ###### _Example JSON config for StartCardsModifiedRule_
+  <details>
```
jimconner, line 557: If we are changing to use <details> blocks in the docs, it would be good to ensure that we caputre all of the rules. The more recently added rules for AbilityBackstabAdjusted, BackStabConfig, RegainAbilityifMaxxedOut and  CardClassRestrictionOverriddenRule still have their JSON examples up-front. My personal prefrence is to have the JSON displayed without any extra clicking, but as I suffer from 'interface designed by an engineer syndrome', I feel that Mex should make the final call on the format of the docs.

Buruko: Ugh. I think I'm making it worse not better. By deleting the file from this branch does that mean it will remove it from main?

Buruko:
My play testing shows these work, but I need more feedback before making additional adjustments. So working and need you to scope for errors.

orendain (approved):
As far as these current changes go, the code LGTM!  Whether you want to extend further, or commit as is for now, is totally up to you.

PR #234, opened 2022-03-12
Two more example JSON rulesets to help with documentation etc.

jimconner:
# Two more example rulesets to help people get started writing their own.

## Better Sorcerer ( #STS Save the Sorcerer🪄 )
Applies 0 action cost for the ⚡Zap ability. This very small change 'fixes' the Sorcerer character and makes it much more enjoyable to play in the early stages of the game. Normally the Sorcerer is a bit weak until the card energy meter starts to fill up more rapidly from the 2nd floor down, but having a free zap per round leaves him free to add in a 2hp Melee attack giving an equivalent 3HP attack plus a move like all of the other characters can do.

## 3x3 Potions and Buffs 💪
Applies 3x3 Area of Effect on most buffs. Players near the bard when he sings will all benefit from his songs. If your friends are nearby and thirsty then of course you'd share your drink. Applies to Courage, RepairArmor, Strength, Speed, AntiToxin, Adamant and Heal.

### DIFF
--- a/docs/rulesets/Better Sorcerer.json
+++ b/docs/rulesets/Better Sorcerer.json
@@ -1,0 +1,12 @@
+{
+    "Name": "Better Sorcerer",
+    "Description": "0 Action Cost for Sorcerer's Zap - No other changes. #STS",
+    "Rules": [
+        {
+            "Rule": "AbilityActionCostAdjusted",
+            "Config": {
+              "Zap": false,
+            }
+          },
+    ]
+  }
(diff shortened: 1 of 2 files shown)


docs/rulesets/Better Sorcerer.json
```diff
@@ -0,0 +1,12 @@
+{
+    "Name": "Better Sorcerer",
+    "Description": "0 Action Cost for Sorcerer's Zap - No other changes. #STS",
```
orendain, line 3: > #STS

😆

orendain (approved):
LGTM.  Only question I had was about file naming (including spaces), but I see it's 1-to-1 the name of the ruleset, which makes sense.

PR #224, opened 2022-03-09
Add HouseRules UI near the table in the hangouts space.

orendain:
Resolves #222.

Builds on https://github.com/orendain/DemeoMods/pull/235 to add the HouseRules UI to Hangouts.

Once I got the UI to a place where it wasn't in the way, I stopped moving it.  There are no assurances whatsoever that its placement, or the fact that it auto-faces the user, was given much thought 🙃.  Any suggestions or follow-up are welcome.  

@jimconner had mentioned placing it against one of the bookshelves?

> Note: Should be merged after #235 .

Issue #222, opened 2022-03-09
Consider adding HouseRules UI to the hangouts space.


orendain:
Perhaps next to the table where players congregate to start a game.

PR #235, opened 2022-03-12
Enable UiHelper to render buttons in Hangout spaces.

orendain:
Part of https://github.com/orendain/DemeoMods/issues/222.

There were non-trivial differences in how interactables are created in the Hangouts space.  For buttons, their behavior must be registered with a particular button handler.  Additionally, interactables in Hangouts and the typical VR space operate on two different Unity "layers" when it comes to collisions with the pointer (i.e., the player's hand beam).

**Changes:**
- The `Common` package now includes a patch allowing it to capture a Hangouts-related handler from Demeo.
- As a result of the above, the `Common` package must now be initialized (with `CommonModule.Initialize()`).  This should be called from a dependent mod's `OnApplicationStart()`.
- Added the appropriate `CommonModule.Initialize()` calls to both RoomFinder and HouseRules.
- Removed a complicated check (`UiHelper` checking `DemeoResource.EnsureResourceExists()` which sometimes re-initializes `DemeoResource` values).  This is replaced by dropping the check and always re-initializing `DemeoResource`.  A bit more computational overhead, but reduced complexity and more deterministic.
- Introduced `GetCurrentUiType()`, with corresponding `UiType` enum, which will for now return either `UiType.Vr` or `UiType.Hangouts`.
- Added logic for creating a functional button in Demeo Hangouts (`CreateHangoutsButton()`). This logic, however, is abstracted behind the single `UiHelper.CreateButton()` method.  The method will create a button for whichever UI type (e.g., hangouts or typical VR UI) is currently loaded.

**Known issues:**
- Buttons rendered in Hangouts do not properly get their hover effect applied with this change.

> Note: This change in and of itself does not add RoomFinder or HouseRules UIs to the Hangouts space.  It only enables that feature.  See https://github.com/orendain/DemeoMods/pull/224 for the PR that builds on this and adds HouseRules UI to Hangouts.

### DIFF
--- a/Common/Common.csproj
+++ b/Common/Common.csproj
@@ -52,6 +52,7 @@
         </Reference>
     </ItemGroup>
     <ItemGroup>
+        <Compile Include="ModPatcher.cs" />
         <Compile Include="Properties\AssemblyInfo.cs" />
         <Compile Include="CommonModule.cs" />
         <Compile Include="UI\DemeoResource.cs" />
--- a/Common/CommonModule.cs
+++ b/Common/CommonModule.cs
@@ -1,9 +1,24 @@
 ﻿namespace Common
 {
+    using Bowser.Core;
     using MelonLoader;
 
-    internal class CommonModule
+    internal static class CommonModule
     {
         internal static readonly MelonLogger.Instance Logger = new MelonLogger.Instance("Common");
+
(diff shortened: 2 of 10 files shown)


jimconner (approved):
LGTM.

PR #237, opened 2022-03-12
Move board syncing functionality to its own class.

orendain:
Encapsulates board syncing logic in its own class.  No new functionality added.

Restructure flow a bit to allow for subsequent PRs that extend syncing functionality.

Also sets style SA1202 severity to none.  Always wondered if the style rule would be practical, but hasn't been.


orendain:
Moving to draft as some more changes are being made.

jimconner (approved):
I'm liking the direction you've gone in with BoardSyncer. Much better :-)

PR #238, opened 2022-03-13
Resync board only at specific times, rather than as soon as a syncable change occurs.

orendain:
**Changes:**
- Trigger sync only at specific opportune times, rather than as soon as a syncable change happens.  Specically, sync immediately if needed during a player's turn, and only at the end of a monster's turn.
- Add minibarricade ability as a new spawn trigger.
- Remove `GoToNextLevel` event as a trigger for new spawns.  No idea why it was added in the first place.
- Apply some small name changes.


HouseRules_Core/BoardSyncer.cs
```diff
@@ -97,6 +95,7 @@ private static bool CanAbilityEventRepresentNewSpawn(SerializableEventOnAbilityU
                 case AbilityKey.RaiseRoots:
                 case AbilityKey.CallCompanion:
                 case AbilityKey.DigRatsNest:
+                case AbilityKey.MiniBarricade:
```
jimconner, line 98: If we're doing MiniBarricade, should we also be doing MagicWall ?

jimconner (approved):

HouseRules_Core/BoardSyncer.cs
orendain: Great catch! Fixed.

PR #240, opened 2022-03-13
Add an enum, representing specially synced data, that rules can declare they are modifying.

orendain:
**Issue:**

Even if we know at what all points we need to sync and are able to have properly-timed syncing strategies to reduce network/computational overhead, we are still resyncing despite not knowing if any active rulesets require it.

**Solution:**

Have rules declare what, if any, special data they may make modifications to.  Declaring this is optional, and defaults to rules not making modification to known "special sync data".

A subsequent PR will use these flags in `BoardSyncer.cs` to reduce even further when we trigger resyncs.

---

I have added flags to three rules:
- StatusEffectConfigRule
- PieceImmunityListAdjustedRule
- PieceConfigAdjustedRule

It my current tired state, I am unaware if any other rules make changes to either Piece data, status effect immunities, or status effect data.  @jimconner any I might be missing?


jimconner (approved):
"It my current tired state, I am unaware if any other rules make changes to either Piece data, status effect immunities, or status effect data. @jimconner any I might be missing?"

I my awake state, I think you've covered all the cases here... At least, all of the cases which are immediately obvious to me.

PR #242, opened 2022-03-13
Trigger board sync only if active ruleset requires it.

orendain:
Ruleset class now exposes the aggregate data type that it makes modifications to.

Board will only sync if there is overlap between what the ruleset needs and what was changed.


jimconner (approved):
All looks great to me 👍

PR #247, opened 2022-03-19
Fix certain replaced values not being restored on rule deactivation.

orendain:
Introducing these changes separate from the `1.13` update effort as these aren't related to changes in Demeo.

**Changes:**

- Fix retrieving piece DTO arrays as lists.
  - In several rules, the issue was that _arrays_ were being retrieved as `Lists` via `Traverse...<List<...`.
  - Instead, they are now retrieved directly.
- Reduce code duplication in two rules by introducing a helper.
  - The helper is nothing new.  The issue was that originally, `OnDeactive` was looping over the collection `_originals`.  However, `_originals` was then being again within the loop.  Modifying a collection while iterating over it often causes issues (because the next element to loop over is often no longer there).
  - Replacing with a helper method inherently solves this issue as the collection modified is not the one being iterated over.

**Testing:** _(For `StatModifiersOverriden` and `RegainAbilityIfMaxxedOutOverridden`)_

Sorcerer starts with 2x Strength and 2x Swiftness.  On first use, each should provide `+3`.  On second use, no effect and card should not be regained.

```json
{
  "Name": "Testing",
  "Description": "Testing",
  "Rules": [
    {
      "Rule": "StatModifiersOverriden",
      "Config": {
          "Speed": 3,
          "Strength": 3
      }
    },
    {
      "Rule": "RegainAbilityIfMaxxedOutOverridden",
      "Config": {
          "Speed": false,
          "Strength": false
      }
    },
    {
      "Rule": "StartCardsModified",
      "Config": {
        "HeroSorcerer": [
          {
              "Card": "StrengthPotion",
              "IsReplenishable": false
          },
          {
              "Card": "StrengthPotion",
              "IsReplenishable": false
          },
          {
              "Card": "SwiftnessPotion",
              "IsReplenishable": false
          },
          {
              "Card": "SwiftnessPotion",
              "IsReplenishable": false
          }
        ]
      }
    }
  ]
}
```

**Testing:** _(For all other rules)_

Simple method:

- Start game with Arachnophobia, and end game.
- `Piece*` rules should deactivate without issue.

More complicated method:

- Load up Arachnophobia ruleset as it modifies all but one of the `Piece*` rules.
- During gameplay, notice rules have taken effect (e.g., Heroes are immune to disease/frozen, Heroes start with more health, Spit acid).
- After gameplay, notice changes do not persist in regular games (e.g., Heroes start with regular health, sorcerer is not immune to his own freeze.)


jimconner (approved):
LTGM - No unexpected errors with either Testing.json or Arachnophobia.json rulesets.

PR #248, opened 2022-03-19
Modify ability rules to perform abilityKey-to-ability lookups.

orendain:
**Notes:**
- Added our typical helper method to the rules to de-duplicate code.
- Moved to be called on `OnPreGameCreated`, instead of `OnPostGameCreated`, as:
  1. It now can be.
  2. Fail earlier if need to.
- Throws an error if a specified abilityKey does not map to a valid ability.


jimconner (approved):
Added commits to fix up  rulesets which still used old ability names. LGTM now.. Merging.

PR #249, opened 2022-03-19
Encapsulate processes dealing with directing the lifecycle of a ruleset.

orendain:
In the same vein as #237, this PR moves the directorship of a ruleset (e.g., when to activate, deactivate, trigger the different hooks, etc.) to an encapsulated class.

No new functionality, just moving of procedures and minor renames.

**Testing**:
- Load any ruleset as per usual.
- HouseRules should continue logging when the game runs activates rules, runs `OnPreGameCreated`, `OnPostGameCreated`, and deactivates rules.
- The welcome message should continue to show up as per usual.


jimconner (approved):
Looks good, works good. Merging.

PR #250, opened 2022-03-19
Arachnophobia - Statmodifiers got renamed in error.

jimconner:
These got changed with some global search/replace. Unfortunately the StatModifiers don't share the same names as the corresponding AbilityKey anymore.


orendain (approved):
LGTM.  That reminds me - I'll patch `RegainAbilityIfMaxxedOutOverridden` and whatever remaining rules are still using the `(Clone)` approach.

PR #253, opened 2022-03-20
Paginate HouseRules' Ruleset selection UI.

orendain:
Uses the new `PageStack` feature in the common module to paginate the HouseRules Ruleset selection UI.

Additionally:
- Registers the `None` ruleset so it appears naturally in the list instead of having to be injected.

image


jimconner (approved):
Looks and works splendidly.

PR #256, opened 2022-03-20
Modify ExampleRulesetExporter to export all registered rules.

orendain:
**Changes:**
- Removes the old `ExampleRulesetExporter `, which created and exported a single ruleset.
- Releases dependency on HR Essentials mod.
  - As a result of no longer creating an example ruleset in the configuration mod, it no longer has a dependency on the mod where the rules/rulesets are defined.
- Introduce a new `ExampleRulesetExporter` that exports all registered rulesets as JSON to the path `UserData\HouseRules\ExampleRulesets`.
  - In the process, exported rulesets are prepended with the string `(Custom) `.
- Add filename sanitization usable in general export procedures.

**Note:**

There are two lines in `ExampleRulesetExporter.cs`:
```cs
// Uncomment to export registered rulesets.
// ExportRegisteredRuleset();
```

These lines may be uncommented to view the effects.  This will be part of the release cycle, so that all registered rulesets are written out, but not during regular use.  This may be made easier with the use of `#pragma` and the `DEBUG` (or some other) constant, but is not part of this PR.

### DIFF
--- a/HouseRules_Configuration/ConfigManager.cs
+++ b/HouseRules_Configuration/ConfigManager.cs
@@ -12,7 +12,7 @@
 
     public class ConfigManager
     {
-        private static readonly string RulesetDirectory = Path.Combine(MelonUtils.UserDataDirectory, "HouseRules");
+        internal static readonly string RulesetDirectory = Path.Combine(MelonUtils.UserDataDirectory, "HouseRules");
 
         private readonly MelonPreferences_Category _configCategory;
         private readonly MelonPreferences_Entry<string> _defaultRulesetEntry;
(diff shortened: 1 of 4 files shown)


jimconner (approved):
Works perfectly. I can see ExportedRulesets correctly after I uncomment the line.

jimconner:
Merged, but a question around the settings for exports. Are there any options for NewtonSoft to make it give a more condensed output for easier editing? NewtonSoft seems to really like putting each thing on a separate line.

e.g. this.. 
```
      "Rule": "StartCardsModified",
      "Config": {
        "HeroGuardian": [
          { "Card": "HealingPotion", "IsReplenishable": false },
          { "Card": "ReplenishArmor", "IsReplenishable": true },
          { "Card": "Bone", "IsReplenishable": true },
```
versus...
```
      "Rule": "StartCardsModified",
      "Config": {
        "HeroBard": [
          {
            "Card": "CallCompanion",
            "IsReplenishable": false
          },
          {
            "Card": "SummonElemental",
            "IsReplenishable": false
          },
          {
            "Card": "HuntersMark",
            "IsReplenishable": true
          }
        ],
```

orendain:
Not by default.  Only `None` and `Indented` [formatting options](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Formatting.htm)

[The ability to do so is there](https://stackoverflow.com/questions/66993700/can-i-collapse-ignore-levels-when-deserializing-json), but would require writing custom converters.

PR #257, opened 2022-03-20
Codify JSON rulesets.

orendain:
As per discussion, this PR brings in the current JSON rulesets as coded rulesets.

> Note: I ran out of fuel towards the end and _**did not**_ go back to double-check that all rules were indeed translated over.  While HR builds, I did not test Demeo startup.  I can finalize later, or if someone else does please feel free to extend this branch.

### DIFF
--- a/HouseRules_Essentials/Rulesets/AoePotionsAndBuffsRuleset.cs
+++ b/HouseRules_Essentials/Rulesets/AoePotionsAndBuffsRuleset.cs
@@ -1,0 +1,39 @@
+﻿namespace HouseRules.Essentials.Rulesets
+{
+    using System.Collections.Generic;
+    using DataKeys;
+    using HouseRules.Essentials.Rules;
+    using HouseRules.Types;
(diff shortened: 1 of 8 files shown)


HouseRules_Essentials/Rulesets/AoePotionsAndBuffsRuleset.cs
```diff
@@ -0,0 +1,39 @@
+﻿namespace HouseRules.Essentials.Rulesets
+{
+    using System.Collections.Generic;
+    using DataKeys;
+    using HouseRules.Essentials.Rules;
+    using HouseRules.Types;
+
+    internal static class AoePotionsAndBuffsRuleset
```
orendain, line 8: Note: Named as such because class names can not begin with a number (e.g., `3x3Potions`).  Feel free to rename to anything better.

jimconner (approved):
All looks good to me. Merging.

Issue #260, opened 2022-03-23
mods did NOT trigger from the lounge even though the screen was present


jimconner:
User feedback from @solacd... 

All the adverstised ruleset changes appeared to work properly (AOE, card changes) but I note that the mods did NOT trigger from the lounge even though the screen was present.  I had to exit, restart and run directly from Demeo to get the mods to work.

Apparently mods do not trigger when selected from within the hobby shop. We should investiagte.

orendain:
This could add weight to my earlier (and hopefully incorrect) speculation that the host of a Hangouts game comes down to a race condition on who can reserve a room first.

Explored this a bit tonight, but more investigation needed.

orendain:
Results from further exploration, and update on fix:

In addition to reports starting coming in about HR not working when starting a game from Hangouts, others were saying HR was self-activating when they joined a game in Hangouts where they weren't the host.

After some investigation, we found that the culprit is how Hangout games select the host.  When players start a game from within Hangouts, the person who starts the table isn't automatically the host.  Instead, the host+clients race out of hangouts and into the not-yet-created game room.  First one to successfully create the room becomes host, and all others become regular clients.

Thus, sometimes an HR-modded player starts a modded game inadvertently, and other times they want to host a game but their host seat gets taken.

---

There's no mechanism I could find by which to ensure clients don't steal the host's seat.  Delaying the local player so as to guarantee they don't take host if they don't mean to is the easy half of it.

---

The process for exiting hangouts, reloading the lobby scene, and finally attempting to create a game is a long one.  On my PC, it takes about 5.5 seconds from the moment I send the start signal to other players to get a new room registered.  And about +1 second if it's the first time I've created a game that session.

What I've resorted to is creating a more optimal path from Hangouts to room creation when a HR-player wants to host a game.  E.g.,:
  - I frontload as much of the computation before sending out the signal to players that they should race to the room.  Unmodded clients will do all this as part of the race, while the HR modded player has it already completed.
  - I collapsed what is usually a dozen or so stackframes to less than half.
  - I forego any visual exit/entrance fading effects, and skip corresponding audio effects.

I've been able to shave off about 1 second from my time.  Down from 5.5 seconds to 4.5.  Though, I'm not sure this would be enough for everyone.  But we'll see.

---

We've also considered taking back host once the game begins, but controlling pre-game sequences is pretty important for enabling a lot of features.

PR #267, opened 2022-03-27
Enable HouseRules users to race through the Hangout game starting procedure.

orendain:
Resolves https://github.com/orendain/DemeoMods/issues/260.

A subsequent PR can focus on the converse of this: to _slow_ HouseRules users if they do not want to be the host.

Here are my latest findings (3 trials, results in seconds).  "Warmed" suggests a game was previously started that session, "Speed" suggests this feature was enabled.

TL;DR:  This feature shaves off about a second.

```
Not warmed, no speed:
5.729
5.608
5.82

Warmed, no speed:
5.147
5.637
5.639

Not warmed, speed:
4.284
4.631
4.87

Warmed, speed:
4.698
4.574
4.643
```

> Note:  During speeding through, there may be a message in the log `[UnityExplorer] [WARNING] The target UnityEngine.Object was destroyed!` which I believe is due to skipping the full instantiation of the `SocialProviderPar`.  Not 100% verified, but the warning should be harmless.

### DIFF
--- a/HouseRules_Core/CoreMod.cs
+++ b/HouseRules_Core/CoreMod.cs
@@ -15,6 +15,7 @@
             var harmony = new Harmony("com.orendain.demeomods.houserules.core");
             LifecycleDirector.Patch(harmony);
             BoardSyncer.Patch(harmony);
+            HangoutsGameRacer.Patch(harmony);
 
             HR.Rulebook.Register(Ruleset.None);
         }
--- a/HouseRules_Core/HangoutsGameRacer.cs
+++ b/HouseRules_Core/HangoutsGameRacer.cs
@@ -125,7 +125,7 @@
 
             StopWatch.Stop();
             var timeElapsed = StopWatch.Elapsed;
-            HR.Logger.Msg($"Time to join game from Hangouts: {timeElapsed.Seconds:00}.{timeElapsed.Milliseconds:00}s");
+            HR.Logger.Msg($"[HangoutGameRacer] Time to join game from Hangouts: {timeElapsed.Seconds:00}.{timeElapsed.Milliseconds:00}s");
         }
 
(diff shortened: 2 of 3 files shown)


jimconner:
Should we have a check in here to ensure that the HangoutsGameRacer is enabled only if a ruleset other than 'None' is selected? 

I'm thinking about the case where two players both have HR installed, but only one of those players chose a ruleset to play. Having the ruleset applied if someone selected one feels like the right default behaviour to attempt.

jimconner (approved):
Code all looks good. Merging. The point about disabling Racer when ruleset is None can always be dealt with separately.

orendain:
> Should we have a check in here to ensure that the HangoutsGameRacer is enabled only if a ruleset other than 'None' is selected?

Great point.  I suppose if `None` is selected, then currently this just acts as a "race to become host" feature.  Check ought to be added as a follow-up 👍🏾 .

> I'm thinking about the case where two players both have HR installed, but only one of those players chose a ruleset to play. We should be making best efforts to apply the ruleset applied if someone selected one, so denying the Racer advantage to 'None' seems a reasonable approach.

Hmm, I hadn't thought about this case.  This feature works by hooking into when the person that started the Hangouts table (the "table host") clicks on "Start Game".  Normally, at that point, the table host sends out a start message to everyone (themselves included) that triggers the race.  `HangoutsGameRacer` will hold off on sending the start message until it gets a chance to front-load work that would normally be done post-start-message.  Additionally, it optimizes its post-start-message flow.

`HangoutsGameRacer` won't be able to apply all performance gains to all HR players (unless they're the table host), but we can indeed encapsulate these changes into pre- and post- start message improvements.  So that HR players that aren't table hosts still (hopefully) get some improvements.

Issue #268, opened 2022-03-28
Enable `HangoutsGameRacer` optimizations only when a ruleset is selected.


orendain:
As discussed here: https://github.com/orendain/DemeoMods/pull/267#issuecomment-1080791701

Additionally, encapsulate these changes into pre- and post- start message improvements. So that HR players that aren't table hosts still (hopefully) get some improvements.

PR #269, opened 2022-03-31
Enable HangoutsGameRacer optimizations only when a ruleset is selected and for non-table hosts.

orendain:
Resolves https://github.com/orendain/DemeoMods/issues/268.

**Changes:**
- Only optimize race path when user has a ruleset selected.
- Restructured functions to allow non-table-hosts to receive out-of-hangouts-path optimizations.
- Only display time-to-join when it was being timed.
- Reduce enforced delay before room creation down from 500ms to 50ms.


jimconner (approved):
LGTM - Merging.

PR #275, opened 2022-04-04
Extend LevelSequence rule to allow arbitrarily long sequences.

orendain:
Extend LevelSequence rule to allow arbitrarily long sequences.

Added some notes to explain reasons for certain changes.

### DIFF
--- a/HouseRules_Essentials/HouseRules_Essentials.csproj
+++ b/HouseRules_Essentials/HouseRules_Essentials.csproj
@@ -97,6 +97,9 @@
       <HintPath>$(SteamDemeoDir)\MelonLoader\MelonLoader.dll</HintPath>
     </Reference>
     <Reference Include="System" />
+    <Reference Include="PhotonUnityNetworking">
+      <HintPath>$(SteamDemeoDir)\demeo_Data\Managed\PhotonUnityNetworking.dll</HintPath>
+    </Reference>
     <Reference Include="UnityEngine.CoreModule">
       <HintPath>$(SteamDemeoDir)\demeo_Data\Managed\UnityEngine.CoreModule.dll</HintPath>
     </Reference>
--- a/HouseRules_Essentials/Rules/LevelSequenceOverriddenRule.cs
+++ b/HouseRules_Essentials/Rules/LevelSequenceOverriddenRule.cs
@@ -3,20 +3,14 @@
     using System.Collections.Generic;
     using System.Linq;
     using Boardgame;
-    using Boardgame.LevelLoading;
-    using Boardgame.Networking;
-    using Data.GameData;
-    using DataKeys;
     using HarmonyLib;
     using HouseRules.Types;
-    using UnityEngine;
 
     public sealed class LevelSequenceOverriddenRule : Rule, IConfigWritable<List<string>>, IPatchable, IMultiplayerSafe
     {
         public override string Description => "LevelSequence is overridden";
 
         private readonly List<string> _adjustments;
-        private List<string> _originals;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="LevelSequenceOverriddenRule"/> class.
@@ -25,19 +19,13 @@
         public LevelSequenceOverriddenRule(List<string> adjustments)
         {
             _adjustments = adjustments;
-            _originals = new List<string>();
         }
 
         public List<string> GetConfigObject() => _adjustments;
 
-        protected override void OnPostGameCreated(GameContext gameContext)
+        protected override void OnPreGameCreated(GameContext gameContext)
         {
-            _originals = ReplaceExistingProperties(_adjustments);
-        }
-
-        protected override void OnDeactivate(GameContext gameContext)
-        {
-            ReplaceExistingProperties(_originals);
+            ReplaceExistingProperties(_adjustments, gameContext);
         }
 
         private static void Patch(Harmony harmony)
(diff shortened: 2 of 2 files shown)


HouseRules_Essentials/HouseRules_Essentials.csproj
```diff
@@ -97,6 +97,9 @@
       <HintPath>$(SteamDemeoDir)\MelonLoader\MelonLoader.dll</HintPath>
     </Reference>
     <Reference Include="System" />
+    <Reference Include="PhotonUnityNetworking">
+      <HintPath>$(SteamDemeoDir)\demeo_Data\Managed\PhotonUnityNetworking.dll</HintPath>
+    </Reference>
```
orendain, line 102: Turns out that it is because when making the following call
```cs
Traverse.Create(gameContext.gameStateMachine).Field<LevelSequence>("levelSequence").Value;
```

`gameStateMachine` (from `GameStateMachine.cs`) contains references to objects from this library.

HouseRules_Essentials/Rules/LevelSequenceOverriddenRule.cs
```diff
@@ -27,37 +19,53 @@ public sealed class LevelSequenceOverriddenRule : Rule, IConfigWritable<List<str
...
+                original: AccessTools.Method(
+                    typeof(LevelSequenceConfiguration), "GetSequenceDefinition"),
+                prefix: new HarmonyMethod(typeof(LevelSequenceOverriddenRule), nameof(LevelSequenceConfiguration_GetSequenceDefinition_Prefix)));
         }
 
-        protected override void OnDeactivate(GameContext gameContext)
+        /// <remarks>
+        /// Sets a safe sequence definition even if the active game type does not have one that extends to the current level.
+        /// </remarks>
+        private static bool LevelSequenceConfiguration_GetSequenceDefinition_Prefix(ref SequenceDefinition __result, int index, LevelSequence.GameType gameType)
```
orendain, line 42: `BoardGameActionGoToNextLevel.StartNextLevel()` was calling `levelManager.GetCurrentLevelObject()` which was calling `levelSequenceConfiguration.GetSequenceDefinition()` which would throw an exception if the current level index was out of bounds of what the predefined gametype had defined.

This method patches `levelSequenceConfiguration.GetSequenceDefinition()` so that if the level index is out of bounds, we return some sequence definition within bounds (the last sequencedefinition if the last level, or the 3rd-to-last definition otherwise).

---

At first glance, the only real purpose of `levelSequenceConfiguration.GetSequenceDefinition()` in `BoardGameActionGoToNextLevel.StartNextLevel()` is to know whether or not the level definition is a shop stage.  However, shop levels seem to load fine even if the returned levelSequence is almost arbitrary (as I've done here).

In case we experience issues with loading shops as part of this rule, chances are we need to re-patch this method more appropriately.  But seems fine thus far.

HouseRules_Essentials/Rules/LevelSequenceOverriddenRule.cs
```diff
@@ -27,37 +19,53 @@ public sealed class LevelSequenceOverriddenRule : Rule, IConfigWritable<List<str
...
-                originals[0], // We need the Entrance to match the current one or the game will crash when moving to the next level.
-            };
-            newlist.AddRange(replacements); // Append user supplied levels
-            Traverse.Create(levelSequence).Field<string[]>("levels").Value = newlist.ToArray();
-            return originals.ToList().GetRange(1, 5);
+            var gsmLevelSequence = Traverse.Create(gameContext.gameStateMachine).Field<LevelSequence>("levelSequence").Value;
+            var originalSequence = Traverse.Create(gsmLevelSequence).Field<string[]>("levels").Value;
+            Traverse.Create(gsmLevelSequence).Field<string[]>("levels").Value =
+                replacements.Prepend(originalSequence[0]).ToArray();
+            return originalSequence.ToList();
```
orendain, line 68: After some investigation, I found that the `levelSequence` from `levelManager` is different than the `levelSequence` from `gameContext`.

`levelManager.levelSequence` is set as part of the game creation procedure.

`gameContext.levelSequence` is more of a temporary field for the purposes of remembering which implied levelSequence the user selected to play.  This is the value that is then passed into `levelManager.levelSequence` during game creation.

By changing `OnPostGameCreated` to `OnPre...`, we could set `gameContext.levelSequence` directly.  This also allows us to get around the "first-level-must-match-game-type" limitation.  For example, I set RK entrance when playing EQ campaigns.

However, one limitation that still applies is that _some_ entrance must be the first level, even if it doesn't match the campaign.  As such, I've left the original "keep-the-first-entrance" logic, but did adjust this so as to apply during `OnPre...`.

---

Additionally, one issue I encountered was that the game would seem to crash at the end of the game.  While not 100% debugged, the issue seemed to be that we were replacing back the original values after deactivation (as we do with every other rule, of course).

It seems like the game didn't like that (perhaps because post-game the level sequence is supposed to be cleared?  I don't know - I didn't really dig).  Either way, _not replacing_ after deactivation fixed this.  Which means not only less code, but seems safe to do, as the levelSequence is discarded after end of game anyhow.  And changing level sequence _during_ matches once they've begun if the host leaves might have unintended consequences.

jimconner: Well done for cracking this one. The LevelSequence from GSM was returning valid level sequences for me, but changes were not getting applied. Thanks for investigating and fixing.

jimconner (approved):

PR #282, opened 2022-04-10
Skip intro loading sequence on PC editions.

orendain:
**Changes:**
- Skip intro loading sequence on PC editions.
- Bump version to `1.4.0`.
- Add the PC edition as a compatible target.

> Note: I would have wanted to more elegantly modify the boot sequence code to skip the starting video, but ran into issues adding a transpiler.


jimconner (approved):
Works perfectly.. Merging.

PR #291, opened 2022-04-14
New rule for controlling the duration of TileEffects.

jimconner:
There are five TileEffects in the game - Gas, Acid, Web, Water and Target(unused). 

The durations for each of these are defined in the GetLayerExpireTime function.

This rule overrides that function so that it can accept a user-defined list of expiry times instead.

### DIFF
--- a/HouseRules_Essentials/Rules/TileEffectDurationOverriddenRule.cs
+++ b/HouseRules_Essentials/Rules/TileEffectDurationOverriddenRule.cs
@@ -41,23 +41,22 @@
                     nameof(TileEffectState_GetLayerExpireTime_Prefix)));
         }
 
-        private static bool TileEffectState_GetLayerExpireTime_Prefix(ref TileEffect type, ref int __result)
+        private static bool TileEffectState_GetLayerExpireTime_Prefix(TileEffect type, ref int __result)
         {
             if (!_isActivated)
             {
                 return true;
             }
 
-            if (_globalAdjustments.ContainsKey(type))
+            if (!_globalAdjustments.ContainsKey(type))
             {
-                __result = _globalAdjustments[type];
+                return true;
             }
             else
             {
-                __result = 9999;
+                __result = _globalAdjustments[type];
+                return false; // We returned an user-adjusted config.
             }
-
-            return false; // We returned an user-adjusted config.
         }
     }
 }
--- a/HouseRules_Essentials/Rules/TileEffectDurationOverriddenRule.cs
+++ b/HouseRules_Essentials/Rules/TileEffectDurationOverriddenRule.cs
@@ -1,0 +1,63 @@
+﻿namespace HouseRules.Essentials.Rules
+{
+    using System.Collections.Generic;
+    using Boardgame;
+    using Boardgame.Board;
+    using DataKeys;
+    using HarmonyLib;
+    using HouseRules.Types;
+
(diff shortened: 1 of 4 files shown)


HouseRules_Essentials/Rules/TileEffectDurationOverriddenRule.cs
```diff
@@ -0,0 +1,63 @@
...
+    using HouseRules.Types;
+
+    public sealed class TileEffectDurationOverriddenRule : Rule, IConfigWritable<Dictionary<TileEffect, int>>, IPatchable, IMultiplayerSafe
+    {
+        public override string Description => "Tile Effect durations are overridden.";
+
+        private static Dictionary<TileEffect, int> _globalAdjustments;
+        private static bool _isActivated;
+
+        private readonly Dictionary<TileEffect, int> _adjustments;
```
orendain, line 17: Woot! Static typing (`TileEffect`) ❤️ .

HouseRules_Essentials/Rules/TileEffectDurationOverriddenRule.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            if (_globalAdjustments.ContainsKey(type))
+            {
+                __result = _globalAdjustments[type];
+            }
+            else
+            {
+                __result = 9999;
+            }
```
orendain, line 58: Is the defaulting to 9999 intentional? Or was the vision to ideally override only those effects that were specified?

orendain (approved):
Just one quick question that could be answered later, but LGTM.

HouseRules_Essentials/Rules/TileEffectDurationOverriddenRule.cs
orendain: If the latter, we could prefix this block with:
```cs
if(!_globalAdjustments.ContainsKey(type)) {
    return true;
}
```

HouseRules_Essentials/Rules/TileEffectDurationOverriddenRule.cs
```diff
@@ -0,0 +1,63 @@
...
+        private static void Patch(Harmony harmony)
+        {
+            harmony.Patch(
+                original: AccessTools.Method(typeof(TileEffectState), "GetLayerExpireTime"),
+                prefix: new HarmonyMethod(
+                    typeof(TileEffectDurationOverriddenRule),
+                    nameof(TileEffectState_GetLayerExpireTime_Prefix)));
+        }
+
+        private static bool TileEffectState_GetLayerExpireTime_Prefix(ref TileEffect type, ref int __result)
```
orendain, line 44: nit: could drop `ref` from `TileEffect` since the ref feature goes unused.

HouseRules_Essentials/Rules/TileEffectDurationOverriddenRule.cs
jimconner: Defaulting to 9999 was what the original code did, so I kept the same. Returning earlier would probably have been a better approach as then you'd only need to specify the overrides... Will fix.

HouseRules_Essentials/Rules/TileEffectDurationOverriddenRule.cs
jimconner: Without the ref, it was sad because I wasn't passing the 'type' parameter into the function. Ref seemed to make it all happy again.

PR #292, opened 2022-04-14
Online version check

jimconner:
This is better than nothing, but not by much!

There's enough functional code in here to report all of the DLL names and versions, fetch the list of releases from github, check whether the currently installed one matches the latest, and display a 'Legacy' tag on the HouseRules menu title.

We should really be doing symantic version checking to see if the release is newer/older than the installed one, but currently it's just checking whether the installed version matches the latest release. This will mean that our own under-development future-versions will be flagged up as 'Legacy'. I did find a code example for symantic version checking (https://gist.github.com/yadyn/959467) that we could use if we want to go down this route. I didn't go down this route for now because I'm not sure that we're using tags/versions properly in AssemblyInfo or github and right now it involves lots of cutting substrings that we might change in the near future. 

I'm also sure we can find some better ways to highlight in the GUI that a newer version is available. It would be nice to add a banner across the corner of the menu or something like that, but that's way beyond my abilities.

During testing, you will need of course need to install the octokit Nuget dependency and copy the octokit.dll from your build dir into UserLibs for the github check to work.

### DIFF
--- a/HouseRules_Configuration/ConfigurationMod.cs
+++ b/HouseRules_Configuration/ConfigurationMod.cs
@@ -24,7 +24,13 @@
                 if (release.Name.StartsWith("HouseRules"))
                 {
                     Logger.Warning($"Latest HouseRules Release {release.Name} has the tag {release.TagName}");
-                    break;
+                    Logger.Warning($"My version {HR.HouseRulesVersion}");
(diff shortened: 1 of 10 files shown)


orendain (approved):
Made the changes described in https://github.com/orendain/DemeoMods/pull/295 and merging per offline conversation.

PR #294, opened 2022-04-15
Use headers to allow anchor links to rules

Thynix:
This allows linking directly to them within the page at the cost of the emphasis of the rule name.


Thynix:
(The force push was to fix the "italicizing" -> "emphasis" in the commit message.)

orendain (approved):
Thanks @Thynix!

PR #316, opened 2022-04-26
Typofix - ScheduleResync method got renamed.

jimconner:
I just pulled main and tried to build, but failed.

This PR fixes that.


orendain (approved):
Thank you @jimconner !

PR #318, opened 2022-04-27
Replace flawed AbilityDamageAdjusted rule with a better alternative.

jimconner:
The old rule did not disable correctly failing to unset the targetDamage and critDamage correctly, and it was not possble to specify critDamage independently.

This new rule allows setting both values and correctly stores and restores default settings.

Also in this commit, updates to the rulesets which previously used AbilityDamageAdjusted


orendain (approved):
Woot, great work!

Issue #296, opened 2022-04-16
Feature idea: card replenishing that takes more than one turn


Thynix:
In making balance changes, it'd be nice to be able to have a little more nuance to whether a card is replenishable:

For example, if I were to make fireball replenishable, but found it to be overpowered, I could reduce its AoE, but it'd be nice to have the option of putting a wait of a few turns between replenishments. That way it would still be a guaranteed part of the kit, but without turning sorcerer into a fireball machine gun.

orendain:
After taking a look at the game code for this:

There's a `IsReplenishing` flag (in `Inventory.cs` for anyone curious), that controls whether a card should be replenished the next turn or not.  It's a boolean (yes/no), so there's no built-in way to adjust the number of rounds it takes to replenish a card.

However, we might be able to write a rule that keeps track of all replenishable cards, and overrides only triggers the appropriate `Inventory.RestoreReplenishables()` method when some specified number of turns have elapsed.

Just some preliminary findings for anyone that might want to attempt it - I may not be able to visit this for at least a week.

jimconner:
Thinking about this one a little, I've had a couple of different ideas that might be viable...

It looks as both IsReplenishable and IsReplenishing are being handled as binary operations on the `flags` integer. IsReplenishable is the lowest bit, and IsReplenishing is bit 1. Only the lower two bits of this `flags` int ever seem to be getting used, so we could potentially use a few of the unsed bits  to hold extra state. We'd have to patch both `TryAddAbilityToInventory` to set some extra bits and `IsReplenishing` to handle a countdown, but we wouldn't have to store any extra information outside of the existing inventory structure. (Is it even possible Inventory.cs to have an extra TryAddAbilityToInventory method that would handle more than a bool for IsReplenishing?)

My second idea was that we could use one of the EffectStates to revent replenishing (in the same way as happens for Sneak). For example, casting a fireball could leave you weakened for 1 round afterwards, and then replenishing could be disabled during Weakend state by patching RestoreReplenishables. This has the disadvantage that the same rule would apply to all players, so anyone who was weakened wouldn't get their items replenished - Could make for an iteresting gameplay dynamic though.

orendain:
Great findings @jimconner, piggybacking on `flags` is super clever!

The second idea seems like it has potential to be a very interesting rule/ruleset. Feels like a "disarm" feature.

jimconner:
Resolved by #319

PR #319, opened 2022-04-30
Store replenishFrequency in the unused bits of Inventory 'flags'

jimconner:
# Allow card replenish frequencies to be user-configured.

This PR is in response to #296. **Feature idea: card replenishing that takes more than one turn**

:construction:**This is a breaking change for any JSON rulesets using StartCardsModified**:construction: due to the change from `bool` to `int` for IsReplenishable within the `CardConfig` struct.

This PR makes use of six of the unused bits in the `flags` integer register in order to store 3-bit values for `refreshFrequency` and `countdown`.
 
// flag bits
// 0 : isReplenishable
// 1 : isReplenishing
// 2-4 : ReplenishCounter - 3-bit range used by RestoreReplenishables for counting rounds.
// 5-7 : ReplenishFrequency - 3-bit number, user-configured target.

I wanted to get this PR'd early to get some fresh reviewer eyes suggesting improvements. 

I made changes to the `SetInventory` method so that instead of calling `piece.TryAddAbilityToInventory` we manipulate `piece.inventory.Items` directly, allowing us to set the flags to include the refreshFrequency target in bits 5-7. However, I'm unable to then set `piece.inventory.needsSync = true` as it's `readOnly` - This may mean that I've broken things for multiplayer (which I haven't tested). This needs testing, and if I have broken things then some advice on how to set the readonly piece.inventory.needsSync would be appreciated.

I prefix patched `RestoreReplenishables` so that it could handle bitwise operations for the `ReplenishCounter` and `ReplenishFrequency` bits. The logic in here is somewhat convoluted and when combined with all of the bitwise operations it is pretty confusing to follow. Amazingly it does all appear to be working correctly (in the limited testing I've done). I'm pretty sure there will be some room for improvement here.

During my testing, I was tweaking the Arachnophobia ruleset's Hunter so that his two different Arrow types had different refresh intervals. I don't _think_ that I've broken 'Sneak', but I also haven't confirmed that. 

## Suggested JSON for testing
 ```json
  {
    "Rule": "StartCardsModified",
    "Config": {
      "HeroHunter": [
        { "Card": "HealingPotion", "IsReplenishable": 0 },
        { "Card": "Arrow", "IsReplenishable": 1 },
        { "Card": "PoisonedTip", "IsReplenishable": 2 },
        { "Card": "CoinFlip", "IsReplenishable": 0 },
        { "Card": "DropChest", "IsReplenishable": 0 },
      ],
    }
  },
  ```

### DIFF
--- a/HouseRules_Essentials/README.md
+++ b/HouseRules_Essentials/README.md
@@ -716,7 +716,8 @@ The [Settings Reference](../docs/SettingsReference.md) contains lists of all dif
   - To configure:
     - Specify the [BoardPieceId](../docs/SettingsReference.md#boardpieceids) of the piece to replace the starting card hand of.
     - Specify the [AbilityKey](../docs/SettingsReference.md#boardpieceids) of the cards to add to the piece's hand.
-    - Specify `true` if the card should replenish, or `false` if it should not.
+    - Specify integer value for replenish frequency. 0=NoReplenish, 1=EveryTurn, 2=Every 2 Turns etc
+    - Max replenish frequency value is 7 
 
   ###### _Example JSON config for StartCardsModified_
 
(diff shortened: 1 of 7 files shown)


jimconner:
Sadly it appears that this only works for the host. I guess clients handle their own replenishes internally :-(

HouseRules_Essentials/Rules/StartCardsModifiedRule.cs
```diff
@@ -47,6 +47,50 @@ private static void Patch(Harmony harmony)
...
                     typeof(StartCardsModifiedRule),
                     nameof(Piece_CreatePiece_Postfix)));
+            harmony.Patch(
+                original: AccessTools.Method(typeof(Inventory), "RestoreReplenishables"),
+                prefix: new HarmonyMethod(
+                    typeof(StartCardsModifiedRule),
+                    nameof(Inventory_RestoreReplenishables_Prefix)));
+        }
+
+        private static bool Inventory_RestoreReplenishables_Prefix(bool __result, Piece piece)
```
orendain, line 57: nit: we'll need the `ref` modifier to propagate changes to `__result`.

HouseRules_Essentials/Rules/StartCardsModifiedRule.cs
```diff
@@ -47,6 +47,50 @@ private static void Patch(Harmony harmony)
...
+            {
+                return true;
+            }
+
+            __result = false;
+            for (int i = 0; i < piece.inventory.Items.Count; i++)
+            {
+                Inventory.Item value = piece.inventory.Items[i];
+                var targetRefresh = (value.flags & 224) >> 5;
+                var countdown = (value.flags & 28) >> 2;
```
orendain, line 69: nit: shifting first and then applying a simpler mask (7, or binary literal `0b111`), or applying mask without shifting (e.g., `0b11100`) may be easier to read.

orendain (commented):
Nice!  I know this is a preliminary PR, so I threw it only a handful of notes as they came up.  Apologies if they were to be addressed eventually and my comments are just noise.

>  unable to then set `piece.inventory.needsSync` = true as it's readOnly

Try this:
```cs
Traverse.Create(piece.inventory).Property<bool>("needSync").Value = true;
```

> Sadly it appears that this only works for the host. I guess clients handle their own replenishes internally :-(

Aww, shucks!  Hmmm... I think your setting of `needSync` may be the solution here.  Since `flags` is part of the payload to clients during an inventory sync, I would expect clients get their flags overwritten.

Now whether or not this sync happens at the time it ought to happen (before the user gets a chance to user their cards) is a different question 🤔.

---

Will fire up this change and see if I can test the sync timings above.

orendain:
After setting `needSync`, this rule seems to work as expected for my very simple 2-player multiplayer setup.

To get the host's inventory to update properly each round, modifying `bool __result` with `ref` should do the trick.

jimconner:
Nice - Thanks for getting it working @orendain - I've added an extra commit to address all of the nits and implement fixes as required. All appears to work correctly in my simple 2-player setup now too.

HouseRules_Essentials/README.md
```diff
@@ -725,28 +726,28 @@ The [Settings Reference](../docs/SettingsReference.md) contains lists of all dif
...
     "Config": {
       "HeroGuardian": [
-        { "Card": "HealingPotion", "IsReplenishable": false },
-        { "Card": "ReplenishArmor", "IsReplenishable": true },
-        { "Card": "WhirlwindAttack", "IsReplenishable": true },
-        { "Card": "PiercingThrow", "IsReplenishable": false },
-        { "Card": "CoinFlip", "IsReplenishable": false },
-        { "Card": "TheBehemoth", "IsReplenishable": false },
-        { "Card": "SwordOfAvalon", "IsReplenishable": false },
+        { "Card": "HealingPotion", "IsReplenishable": 0 },
```
orendain, line 729: Does it make sense to switch from `IsReplenishable` to something less boolean-sounding in a later patch?  E.g., `Cooldown`, `ReplenishableRounds`, etc.?

orendain (approved):
Awesome!  Just 1 comment for future consideration.

PR #320, opened 2022-04-30
Its a trap ruleset

jimconner:
# 💣It's a Trap Ruleset💣

Build fiendiesh traps for your enemies and lure them to their deaths... but do try not to kill your friends. 

A new ruleset for HouseRules to keep things interesting.

* This new ruleset gives all of the players a 0-cost BoobyTrap which replenishes every turn. This is supplemented with plenty of Lamps, MagicBarrier and Vortex cards to allow for large scale chain-reaction trap sequences to be constructed.
* Torch and DetectEnemies/EyeOfAvalon pieces are no longer targetted by enemies and can be used to view enemy movements whilst you're planning your traps.
* Enemies cannot open doors for themselves.. they have to wait until you let them into your trap.
* TileEffect statuses such as Gas , Acid and Web have been extended to 10 rounds to allow greater destruction when traps are finally triggered. 
* The Assassin's Sneak ability has been extended to 5 turns to allow for lamps to be placed deeper into enemy territory - the Sneak card will also periodically get dealt to other players. 
* Players all start with 30HP. Healing, Strength and Speed Potions have 3x3 AoE. AltarsOfHealing are disabled and replaced with HealingBeacons to encourage regrouping.
* Different combinations of lamp-types used for each floor to keep trap construction techniques varied.

After quite a few different revisions, this ruleset has been playing well for the last couple of test games. It is probably ready for a wider audience now. It is certainly ready for other people to be testing it out.

Some notes.

I have only tested with 3-player config. I have tried to make each of the heros work in slightly different ways, but I generally prefer playing certain combos leaving others untested. I find that the ruleset plays well on the Elven map where the doors keep the enemies well contained. My favoured combo is Hunter + Sorcerer + Assassin. I don't think I've played a single game with the Bard in the party.

### DIFF
--- a/HouseRules_Essentials/EssentialsMod.cs
+++ b/HouseRules_Essentials/EssentialsMod.cs
@@ -62,6 +62,7 @@
             HR.Rulebook.Register(LuckyDip.Create());
             HR.Rulebook.Register(TheSwirlRuleset.Create());
             HR.Rulebook.Register(BeatTheClockRuleset.Create());
+            HR.Rulebook.Register(ItsATrapRuleset.Create());
             HR.Rulebook.Register(HuntersParadiseRuleset.Create());
             HR.Rulebook.Register(DifficultyEasyRuleset.Create());
             HR.Rulebook.Register(DifficultyHardRuleset.Create());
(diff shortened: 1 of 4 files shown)


orendain (approved):
LGTM! Can't wait to try this out!

PR #327, opened 2022-05-09
Adjust `ItsATrapRuleset` with updated `StartCardsModifiedRule` format.

orendain:
Caused by not updating the https://github.com/orendain/DemeoMods/pull/319 branch before merging. My fault.

Would have been prompted to to so by GitHub before merge, except that the "Require branches to be up to date before merging" option does not take effect unless there are status checks enabled for the branch.  And there is currently no public CI pipeline (_partly_ blocked by not being allowed to host Demeo DLLs in a public location) to use to run build checks against.


jimconner (approved):
Nice catch - Thanks Mex

PR #330, opened 2022-05-09
Add new rulesets and update others for compatibility with HR1.4

jimconner:
Some of the example rulesets in our docs directory were outdated and wouldn't have worked correctly with the current `main` branch. Some of the newer rulesets were missing.

This fixes that.


orendain (approved):
Woot, LGTM!

PR #335, opened 2022-05-10
Ruleset rebalance tweaks from Gray Alien

jimconner:
Gray Alien's fixes for SpawnCategories so that there are:

- no Wyverns on floor 1
- better chance of chestgoblins so that there can be more loot.. but you have to work for it.


orendain (approved):
LGTM, thanks for porting this into code 🙏🏾 .

TheGrayAlien:
Thanks for all the help Jim. Couldn't have done it without you!

PR #336, opened 2022-05-15
Introduce a new rule: TurnOrderOverridden, allowing the player turn order to be customized.

orendain:
Adds a new rule, `TurnOrderOverridden`, that overrides the player turn order.  It works by simulating the "initiative" stat in popular table-top games, where players with a higher initiative score move first.

**Example JSON config:**

```json
{
  "Rule": "TurnOrderOverridden",
  "Config": {
    "Assassin": 8,
    "Bard": 2,
    "Guardian": 20,
    "Hunter": 6,
    "Sorcerer": 4,
    "Downed": 10,
    "Javelin": 10,
  }
},
```

In the example above:
- When no player is downed or has a Sigataurian Javelin, guardians would go
  first, followed by assassins, then hunters, then sorcerers, and finally bards.
- Downed players will almost certainly go first, except when there's a guardian
  in the group or someone has a Sigataurian Javelin.
- A downed sorcerer with a Sigataurian Javelin will go before a downed hunter
  that doesn't have a Sigataurian Javelin.


orendain:
On second thought ... maybe it would be more prudent to not release this as an independent mod.  Will hold off now until discussed.

orendain:
> Some other ideas that spring to mind:

Nice, will keep those in mind!  This also plays into the triggers/event-based rules previously discussed.

jimconner (approved):
LGTM - Nice work @orendain

PR #337, opened 2022-05-16
2nd Rebalnce of GrayAlien's DemeoReloaded Ruleset

jimconner:
* Tweaked enemy spawns and added Sigatuar chance in case Roots of Evil is played.
* Lowered Mimic melee damage. It was too high, especially when it's almost always Berserk.
* Added more water trail chances to other enemies.
* Increased Chest Goblin melee damage and added an ability and immunities.
* Added extended (25 turns) torches to start cards in case Rat King is played.
* Lowered enemy health increase to 25% from 33%. (needs to be changed in the description)
* Fixed too low gold per floor.


jimconner:
Superseeded by Gray Alien's PR which contains these changes and more.

PR #338, opened 2022-05-16
It's A Trap ruleset tewaks.

jimconner:
A few tweaks after having played It's A Trap! a few times.

* There were far too many chests on the boss level.
* Sorcerer had very limited selection of cards and received too many potions.
* Bard's buff is too overpowered for this ruleset, so reduced it's duration.


orendain (approved):
LGTM!

Issue #323, opened 2022-05-03
Add a rule to include allies and/or friendly monsters when using the Regroup card.


orendain:
The regroup ability makes use of the `Boardgame.GameplayEffects.Teleport` sequence.  The `Teleport` sequence has a field, `teleportPieceMode`, of type `TeleportTargetPiece`, which specifies what pieces can be teleported (`Attacker`, `Defenders`, `MyTeam`).

I suspect overriding `MyTeam` to include allies and/or friendlies would result in them getting pulled in via the regroup card.

---

A related feature could also be:
Overriding `Attacker`/`Defenders` to mean not just a single target, but all nearby attackers/defenders.  This would make banish an AoE banish, and a mystic's teleport an AoE teleport.

jimconner:
Implemented in #340 - Closing this issue.

PR #340, opened 2022-05-20
Allow Regroup to teleport both enemies and players.

jimconner:
This PR patches the `GetPiecesToTeleport` method to allow selection of both players and allies when using the Regroup ability. 
This is a response to #323 as it looked like an interesting one to tackle.

The original GetPiecesToTeleport method takes a maxNum parameter which tells it how many pieces it should select. For some reason this always appears to be set to 4 for Regroups (though I haven't worked out where or why that number is used). To make this a bit more useful for teleporting allies, I've added a maxNum config parameter to the rule. So far I have only tested that with '8' as a setting and confirmed that it teleported more than 4 allies.

The selection criteria for what to select may need improving. I've excluded props such as the Ballista and included permanently charmed things such as elementals (and hopefully Elvyn Hounds). I haven't included temporarily charmed enemies.

I gave this rule one very brief test in a 2-player setup and the client machine didn't see the allies teleport over, so some imrpovements are going to be needed on the board syncing. 

As usual, I've tried to leave in a number of mistakes, errors, bugs and incorrectly named variables to ensure that the reviewer is fully engaged :-)


HouseRules_Essentials/Rules/RegroupAlliesRule.cs
```diff
@@ -0,0 +1,97 @@
...
+                return true;
+            }
+
+            List<Target> list = new List<Target>(maxNum);
+            if (mode == TeleportTargetPiece.Attacker || mode == TeleportTargetPiece.Defenders)
+            {
+                return true;
+            }
+            else if (mode == TeleportTargetPiece.MyTeam)
+            {
```
orendain, line 64: Can remove `else if` to reduce nesting (since it'll absolutely be the case if execution gets to this part).

A huge benefit to the [guard clause](https://refactoring.guru/replace-nested-conditional-with-guard-clauses) that you introduced above.

HouseRules_Essentials/Rules/RegroupAlliesRule.cs
```diff
@@ -0,0 +1,97 @@
...
+                        list.Add(new Target(piece.effectSink));
+                        count++;
+                    }
+
+                    if (type == PieceType.Player && (piece.IsPlayerFaction() ||
+                        piece.HasEffectState(EffectStateType.ConfusedPermanent)) &&
+                        !piece.HasEffectState(EffectStateType.SelfDestruct) &&
+                        !piece.IsPlayer() &&
+                        !piece.IsProp() &&
+                        count < _maxNum)
```
orendain, line 83: We might be able to combine this and the conditional above since they both do very similar things.

Just an idea:

If all of those piece-related predicates in both conditionals were to be moved to some `isTeleportable()` function, we might be able to have something like a single, neat
```
if (count... && piece.networkI... && isTeleportable(piece))
```

HouseRules_Essentials/Rules/RegroupAlliesRule.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            List<Target> list = new List<Target>(maxNum);
+            if (mode == TeleportTargetPiece.Attacker || mode == TeleportTargetPiece.Defenders)
+            {
+                return true;
+            }
+            else if (mode == TeleportTargetPiece.MyTeam)
+            {
+                var attacker = AccessTools.StructFieldRefAccess<AbilityContext, Target>(ref abilityContext, "attacker");
+                int me = attacker.piece.networkID;
```
orendain, line 66: nit: just for clarity, maybe rename `me` -> `attackerId`, or inline the entire `attacker.piece.networkID` in place where `me` is used.

orendain (commented):
Going to try this out the first chance I get!  In the meantime I've added some comments.

HouseRules_Essentials/Rules/RegroupAlliesRule.cs
```diff
@@ -0,0 +1,97 @@
...
+        public override string Description => "Allies are included in Regroup.";
+
+        private static int _maxNum;
+        private static bool _isActivated;
+
+        private readonly int _adjustments;
+
+        private enum TeleportTargetPiece
+        {
+            Attacker,
```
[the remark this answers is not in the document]
jimconner: I was pleasantly surprised to find that this worked. I'd fully been expecting red ink in the melon loader logs for this.

orendain (approved):
LGTM!  Also threw in board syncing during testing and looks great!  Will leave merging to you.

PR #343, opened 2022-05-21
Rebalance Demeo Reloaded ruleset and fix build/compile warnings

TheGrayAlien:
Fixed extra space warnings on build/compile

Adjusted some spawns

Change Shanty to 0 AP

Replace second Shanty with single target heal that can only be used every 3 turns

Fixed special enemies weren't dropping cards/loot as intended

Took away Hunter's 0 AP cost pets and extra move range. Replaced with 1 extra AP.

Was too easy! Increased enemy damage/health. Decreased gold, chests and mana gain


jimconner (approved):
LGTM - Merging.

Issue #262, opened 2022-03-23
Consider a rule that triggers on particular conditions.


orendain:
Random idea for a rule:  Rewards and penalties for certain actions.

E.g.:
- Roll a miss:  take 1 damage.
- Roll a crit: restore 1 HP
- Kill 3 enemies in a single turn:  get the ability to spawn a chest (the spawn chest card)
- Damage a fellow player: become panicked
- etc.

PR #345, opened 2022-05-29
Gain a user-configured ability whenever a critical hit is scored.

jimconner:
This PR is a partial response to #262

I initially had this rule only awarding an extra AP after a critical strike, but realised that the config could be extended to give any card desired, and on a per-hero basis.

As with everything I PR, this has been poorly tested, but I think it should all be OK. I've made an attempt to sync after the inventory gets changed so hopefully this will function as expected in multiplayer too.

This is an extremely frustrating rule to test as you need to be rolling crits... It's amazing how long your can to without rolling one. I managed to complete the 1st floor of sewers without a single crit. 

For my testing, I added the following into one of the rulesets.

```
            var freeActionOnCrit = new FreeAbilityOnCritRule(new Dictionary<BoardPieceId, AbilityKey>
            {
                { BoardPieceId.HeroBard, AbilityKey.OneMoreThing },
                { BoardPieceId.HeroHunter, AbilityKey.PoisonedTip },
                { BoardPieceId.HeroSorcerer, AbilityKey.Fireball },
                { BoardPieceId.HeroGuardian, AbilityKey.Bone },
                { BoardPieceId.HeroRogue, AbilityKey.PoisonBomb },
            });

```

### DIFF
--- a/HouseRules_Essentials/Rules/FreeAbilityOnCritRule.cs
+++ b/HouseRules_Essentials/Rules/FreeAbilityOnCritRule.cs
@@ -1,0 +1,63 @@
+﻿namespace HouseRules.Essentials.Rules
+{
+    using System.Collections.Generic;
+    using Boardgame;
+    using Boardgame.BoardEntities;
+    using Boardgame.BoardEntities.Abilities;
+    using DataKeys;
+    using HarmonyLib;
+    using HouseRules.Types;
+
+    public sealed class FreeAbilityOnCritRule : Rule, IConfigWritable<Dictionary<BoardPieceId, AbilityKey>>, IPatchable, IMultiplayerSafe
+    {
+        public override string Description => "Critical Hit gives free card.";
+
+        private static Dictionary<BoardPieceId, AbilityKey> _globalAdjustments;
+        private static bool _isActivated;
+
+        private readonly Dictionary<BoardPieceId, AbilityKey> _adjustments;
+
+        public FreeAbilityOnCritRule(Dictionary<BoardPieceId, AbilityKey> adjustments)
+        {
+            _adjustments = adjustments;
+        }
+
+        public Dictionary<BoardPieceId, AbilityKey> GetConfigObject() => _adjustments;
+
+        protected override void OnActivate(GameContext gameContext)
+        {
+            _globalAdjustments = _adjustments;
+            _isActivated = true;
+        }
+
+        protected override void OnDeactivate(GameContext gameContext) => _isActivated = false;
+
+        private static void Patch(Harmony harmony)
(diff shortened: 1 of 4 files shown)


HouseRules_Essentials/Rules/FreeAbilityOnCritRule.cs
```diff
@@ -0,0 +1,63 @@
...
+                    typeof(FreeAbilityOnCritRule),
+                    nameof(Ability_GenerateAttackDamage_Prefix)));
+        }
+
+        private static bool Ability_GenerateAttackDamage_Prefix(Piece source, Dice.Outcome diceResult)
+        {
+            if (!_isActivated)
+            {
+                return true;
+            }
```
orendain, line 49: Since this function never requires skipping the original function, we can use `postfix` instead of `prefix`, removing the need to return a boolean.

HouseRules_Essentials/Rules/FreeAbilityOnCritRule.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            if (diceResult == Dice.Outcome.Crit)
+            {
+                if (source.IsPlayer() && _globalAdjustments.ContainsKey(source.boardPieceId))
+                {
+                    source.TryAddAbilityToInventory(_globalAdjustments[source.boardPieceId], isReplenishable: false);
+                    HR.ScheduleBoardSync();
```
orendain, line 56: I figured `TryAddAbilityToInventory(..)` itself would trigger a minimal inventory sync, rendering a need for a special board sync unnecessary.  But it might not happen immediately, so IMO no harm in triggering a manual board sync.

HouseRules_Essentials/Rules/FreeAbilityOnCritRule.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        private static bool Ability_GenerateAttackDamage_Prefix(Piece source, Dice.Outcome diceResult)
+        {
+            if (!_isActivated)
+            {
+                return true;
+            }
+
+            if (diceResult == Dice.Outcome.Crit)
```
orendain, line 51: nit: Can inverse check (into a `!=` that results in a `return`) to have the main logic not be indented (often called guard clauses).

orendain (approved):
> This is an extremely frustrating rule to test as you need to be rolling crits... It's amazing how long your can to without rolling one. I managed to complete the 1st floor of sewers without a single crit.

Haha, indeed!

LGTM - awesome new rule!  Only nits that can easily be part of a maintenance/cleanup chore.

Issue #332, opened 2022-05-10
Can bard buff also heal for 1 HP?


orendain:

jimconner:
Yes it can - #346

PR #346, opened 2022-06-02
Add a user-configurable amout of HP to target each time CourageShanty is cast.

### LINKED ISSUE #322: AbilityDamageOverriddenRule does not allow setting of splashDamage and critSplashDamage
When attempting to adjust AoE abilities such as Whirlwind and Charge, changes applied with `AbilityDamageOverriddenRule` do not work correctly. The card shows the correct values for `targetDamage` and `critDamage` but the damage dealt to enemies is unchanged from the defaults.

Upon investigation it appears that the the AoE attacks for Whirlwind and Charge are dealing `splashDamage` and `critSplashDamage` to enemies. The AbilityDamageOverriddenRule does not currently allow for setting of these fields.

Thanks to The Gray Alien for reporting this issue.

jimconner:
A new rule to allow a user-configurable number of HP to be added to a target when CourageShanty is used.
I took a look into #332 and made a rule to implement it.

I'm wondering once again how we can make this kind of rule more flexible (as I was wondering with FreeAbilityOnCrit). Although this rule adds <int> HP, it would be just as reasonable for it to replenish armor, remove weakness, add strength/speed  or any number of other things that people might want. Maybe we should be implementing a generic method that can set any of the `Stats.Type` parmaters in the same way that we did for pieceConfig... Rules like this could then accept lists of Stat.Types and adjustments for them as config.

I added an `HR.ScheduleBoardSync` here because the clients would probably like to know about their extra HP sooner rather than later, but I'm wondering if it would be better to use `target.piece.effectSink.SyncStat(Stats.Type.Health, target.piece.GetHealth(), 1);` to trigger the sync instead. I tested this out with PC-Flatscreen as a client and it still took about 5 seconds for the newly changed Health stat to appear - I'm not sure if that is any better than a regular Boardsync, so left BoardSync in place for now.

Closes #322

## Testing
Add something like the following into a ruleset and give it a whirl.

```
                new CourageShantyAddsHPRule(3),
```

### DIFF
--- a/HouseRules_Essentials/Rules/CourageShantyAddsHPRule.cs
+++ b/HouseRules_Essentials/Rules/CourageShantyAddsHPRule.cs
@@ -1,0 +1,51 @@
+﻿namespace HouseRules.Essentials.Rules
+{
+    using Boardgame;
+    using Boardgame.BoardEntities;
+    using Boardgame.GameplayEffects;
+    using HarmonyLib;
+    using HouseRules.Types;
+
+    public sealed class CourageShantyAddsHPRule : Rule, IConfigWritable<int>, IPatchable, IMultiplayerSafe
+    {
+        public override string Description => "Courage Shanty also adds HP.";
+
+        private static int _globalAdjustments;
+        private static bool _isActivated;
+
+        private readonly int _adjustments;
+
+        public CourageShantyAddsHPRule(int adjustments)
+        {
+            _adjustments = adjustments;
+        }
+
+        public int GetConfigObject() => _adjustments;
+
+        protected override void OnActivate(GameContext gameContext)
+        {
+            _globalAdjustments = _adjustments;
+            _isActivated = true;
+        }
+
+        protected override void OnDeactivate(GameContext gameContext) => _isActivated = false;
+
+        private static void Patch(Harmony harmony)
+        {
+            harmony.Patch(
+                original: AccessTools.Method(typeof(StrengthenCourage), "UpdateEffectsOnTarget"),
+                postfix: new HarmonyMethod(
+                    typeof(CourageShantyAddsHPRule),
+                    nameof(StrengthenCourager_UpdateEffectsOnTarget_Postfix)));
+        }
+
+        private static void StrengthenCourager_UpdateEffectsOnTarget_Postfix(Target target)
+        {
+            if (_isActivated)
+            {
+                target.piece.effectSink.Heal(_globalAdjustments);
+                HR.ScheduleBoardSync();
+            }
+        }
+    }
+}
--- a/HouseRules_Essentials/EssentialsMod.cs
+++ b/HouseRules_Essentials/EssentialsMod.cs
@@ -23,6 +23,7 @@
             HR.Rulebook.Register(typeof(AbilityActionCostAdjustedRule));
             HR.Rulebook.Register(typeof(AbilityRandomPieceListRule));
             HR.Rulebook.Register(typeof(BackstabConfigOverriddenRule));
+            HR.Rulebook.Register(typeof(CourageShantyAddsHPRule));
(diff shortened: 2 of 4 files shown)


HouseRules_Essentials/Rules/CourageShantyAddsHPRule.cs
```diff
@@ -0,0 +1,51 @@
+﻿namespace HouseRules.Essentials.Rules
+{
+    using Boardgame;
+    using Boardgame.BoardEntities;
+    using Boardgame.GameplayEffects;
+    using HarmonyLib;
+    using HouseRules.Types;
+
+    public sealed class CourageShantyAddsHPRule : Rule, IConfigWritable<int>, IPatchable, IMultiplayerSafe
```
orendain, line 9: nit: `CourageShantyAddsHPRule` -> `CourageShantyAddsHpRule` for pascal casing.  As per https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces.

(except we do break this rule already with our `HR` class)

HouseRules_Essentials/Rules/CourageShantyAddsHPRule.cs
```diff
@@ -0,0 +1,51 @@
...
+                original: AccessTools.Method(typeof(StrengthenCourage), "UpdateEffectsOnTarget"),
+                postfix: new HarmonyMethod(
+                    typeof(CourageShantyAddsHPRule),
+                    nameof(StrengthenCourager_UpdateEffectsOnTarget_Postfix)));
+        }
+
+        private static void StrengthenCourager_UpdateEffectsOnTarget_Postfix(Target target)
+        {
+            if (_isActivated)
+            {
```
orendain, line 45: nit: inverting the statement (i.e. `!_isActivating`) to become a [guard clause](https://refactoring.guru/replace-nested-conditional-with-guard-clauses).

orendain (approved):
Excellent, worked great for me!

> I'm wondering once again how we can make this kind of rule more flexible (as I was wondering with FreeAbilityOnCrit). Although this rule adds HP, it would be just as reasonable for it to replenish armor, remove weakness, add strength/speed or any number of other things that people might want.

Indeed. There could be some overlap between this and https://github.com/orendain/DemeoMods/issues/262. I might be overthinking this, but if "CastingCourageShanty" was a condition, then "HealBy1Health" could be the trigger.  Thinking about it that way, building out a framework extension for https://github.com/orendain/DemeoMods/issues/262 could also bring these kinds of rules under the same umbrella.

> I added an `HR.ScheduleBoardSync` here because the clients would probably like to know about their extra HP sooner rather than later

👍🏾. Since `BoardSyncer` doesn't currently have a `SyncableTrigger` (e.g., `NewPieceModified`, `StatusEffectImmunityModified`, `StatusEffectDataModified`) that picks up on when courage shanty is cast, it might never sync unless told to. 👍🏾 

> I tested this out with PC-Flatscreen as a client and it still took about 5 seconds for the newly changed Health stat to appear

Same here, it seemed to take a few seconds. However, this looks to be due to the delay in the shanty animation.  `BoardSyncer` should be syncing (by way of the logic in `IsSyncOpportunity()` almost immediately during player turns.  Specifically, anytime an outgoing message is sent (e.g., when any player uses courage shanty) it will see the resync request and queue up a resync task.

The timing would be the same if this were replaced/complemented with `SyncStat()`, in that the sync would be queued up to happen immediately.  Of course, `SyncStat()` is much lighter weight (as is the sync inventory equivalent), so that would reduce network load.  With a few rules now relying on full board sync instead of the lighter stat/inventory sync, perhaps we could keep introducing state/inventory-only syncing as part of `BoardSyncer` on our radar if network load gets too bad.

But as far as timing goes, I think each client (host included) is waiting for the shanty animation to finish before a heal task is then processed.

Issue #314, opened 2022-04-25
Think about how rules or rulesets can be limited to specific campaigns or other critera.


orendain:
Following @TheGrayAlien's idea from https://discord.com/channels/841011788195823626/841011980667060244/968149051323338802:

The question is, how can HR restrict a card (via `CardAdditionOverridden` rule) to a particular module/campaign (e.g., RatKing/Forest/Queen)?
- Change `CardAdditionOverridden` so that the applicable campaign can be set.
- Add a new rule, similar to the above, with an extra field for specifying the campaign.

---

A similar request could be made for many of the other rules (e.g., monster spawning, points of interest per floor, etc.).  So the question could then be:  **_should HR have a standard way of building in campaign-filters into each rule?_**

However, there can be multiple filter criteria (e.g., number of active players, campaign type).  Which transforms the question into: **_how can (and what kind of) filter criteria be set for rules?_**
1. (Complex) Maybe in addition to each rule's standard config, specific conditions can be attached to each (e.g., only load a particular rule if playing the RatKing).
2. (Easier) Have conditions that can be attached, but at a ruleset level (e.g., only allow a ruleset to load when RatKing is selected).
3. (Easiest) Do nothing. Encourage rulesets to add a "best played on XYZ map" or "best for 2 players".

---

While point `1.` above may seem like the most robust solution, it may be overly complicated to implement and potentially challenging to elegantly expose those options via JSON for users to customize.

Both points `2.` and `3.` effectively do the same thing: requires rulesets to have campaign-specific variants.

Not sure what the best course of action may be at this point. May need more data.

jimconner:
I don't think that 1. is necessarily too complex. Thinking about how to represent this from a  JSON viewpoint, `Rules` is just a list[] and it is already possible to have the same type of rule specified multiple times (e.g. AbilityAoeAdjusted has cumulative effects if you accidentally have it specified twice) - We could extend the existing format for a rule with an optional `Limit` field on a per-rule basis... 

e.g.
```
    {
      "Rule": "AbilityActionCostAdjusted",
      "Limit": [ "Elven", "Forest" ],
      "Config": {
        "Zap": false,
        "CourageShanty": false,
        "Sneak": false
      }
    },
    {
      "Rule": "AbilityActionCostAdjusted",
      "Limit": [ "Sewers" ],
      "Config": {
        "Torch": true,
        "CourageShanty": true,
        "ReplenishArmor": false
      }
    },
```

[...]

PR #348, opened 2022-06-09
Experimental approach for implementing a way to limit a rule particular adventures.

jimconner:
I was in the mood for playing around with code, so I threw together a way of implementing #314 

This is most definitely a PR that should not be merged - it is purely for discussion.

I selected `LevelPropertiesModifiedRule` for my experiment as it is not an `IPatchable` one and created a `ConfigData` struct to contain both the `Adjustments` and the `Limit`. I then updated all of the rulesets that were using `LevelPropertiesModifiedRule` to enable things to compile. Finally I updated the `ItsATrap!` ruleset so that it would use different properties for Sewers than for others.

The change appears to function successfully, but I don't think this is necessarily the right approach to follow. I'll leave my reasons why on #314 where it can be a part of the discussion.

### DIFF
--- a/HouseRules_Essentials/Rules/LevelPropertiesModifiedRule.cs
+++ b/HouseRules_Essentials/Rules/LevelPropertiesModifiedRule.cs
@@ -7,19 +7,25 @@
     using HarmonyLib;
     using HouseRules.Types;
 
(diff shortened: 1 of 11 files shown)


jimconner:
Closing this PR as it is for discussion only, and it is linked from the issue.

PR #349, opened 2022-06-16
V1.16 compatibility

jimconner:
# Fixup code to work with Demeo 1.16

* The `SearchPiece` enum has had the `SearchPiece.IncludeDisabled` option removed. `SearchPiece.All` is equivalent.
* Turn Queue now initializes to -1 instead of 0 which was causing problems in BoardSyncer when calling `IsPlayersTurn()` - Within the Demeo code `IsPlayersTurn()` went on to call `GetCurrentPiece()` with an index of -1 - This in turn caused a stack trace when accessing the TurnQueue list which can never have a -1 index.
* Updated the reference docs.


orendain (approved):
Woot!  Great finds and fixes!

PR #350, opened 2022-06-17
Document the mod release process.

orendain:
While there are improvements that could be made to our pipelines, this PR documents the current release process.

Special note: There is [a bug with github-changelog-generator](https://github.com/github-changelog-generator/github-changelog-generator/issues/971) that requires a tag exist before generating a changelog.  This means either the changelog is not committed before the tag is created, or that a temporary tag is created just for the purposes of generating a changelog.


RELEASE.md
```diff
@@ -0,0 +1,110 @@
...
+This page describes the release process.
+
+## Prepare for release
+
+### (HouseRules only) Generate JSON for built-in rulesets
+
+Uncomment the relevant line from `ExampleRulesetExporter.cs`. Build and run the
+mod so it generates JSON files for built-in rulesets during game startup.
+
+Open and merge a PR with these JSON files added to `docs/rulesets` for
```
jimconner, line 12: Quite rightly I cannot merge my own PR here.

RELEASE.md
```diff
@@ -0,0 +1,110 @@
...
+
+The `DemeoMods` project uses the
+the [github-changelog-generator](https://github.com/github-changelog-generator/github-changelog-generator)
+for generating the contents of each mod's `CHANGELOG.md`. Each mod directory
+hosts it's own `.github_changelog_generator` config file, as well as its
+generated changelog.
+
+```shell
+cd <mod>
+github-changelog-generator
```
jimconner, line 99: Should be undedrscores here instead of hyphens.

Probably worth mentioning that unless you specify the `--token <personal_access_token>` arguments then you're going to hit the github rate limits pretty quickly.

RELEASE.md
```diff
@@ -0,0 +1,110 @@
...
+
+Every release is tagged with `v<major>.<minor>.<patch>-<mod-name>`,
+e.g. `v1.0.0-houserules`. Note the `v` prefix.
+
+The commit, on the `main` branch, from which to cut a release must have this
+tag. The following example assumes the semantic version "1.0.0" and the mod
+"HouseRules".
+
+```shell
+git tag -s "v1.0.0-houserules" -m "HouseRules v1.0.0"
```
jimconner, line 27: When multiple releases happen simultaneously (HouseRules & Roomfinder for example) I'm guessing that the process is to apply both tags onto the same commit. Guessing is probably not the correct thing to be doing when following documentation, so I'm asking for some clarification to be added here.

RELEASE.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Publish the release
+
+1. Create
+   a [release in GitHub](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository#creating-a-release)
+   using the tag created above.
+2. Include a copy-paste of the generated changelog.
+3. Optionally include a top-level message highlighting the most important points
+   of the release.
+4. Include the bundled ZIP file as a release asset.
```
jimconner, line 110: Curseforge release process?

RELEASE.md
orendain: Fixed.

RELEASE.md
orendain: Thank you!  Fixed.

RELEASE.md
orendain: Good catch - added.

RELEASE.md
orendain: Added clarification👍🏾.

jimconner (approved):

PR #354, opened 2022-06-17
Update `LevelExitLockedUntilAllEnemiesDefeatedRule` to be compatible with Demeo 1.16.

orendain:
The `BoardgameActionPieceDied` constructor had its paramaters updated, which required a corresponding rule change.


jimconner (approved):
LGTM

PR #355, opened 2022-06-17
Remove unneeded logic in `CardLimitModifiedRule` as changes in Demeo 1.16 no longer require it.

orendain:
Remove unneeded logic in `CardLimitModifiedRule` as changes in Demeo 1.16 no longer require it.


jimconner (approved):
Wow.. that's a lot of code cleared away. Nice work :-)

PR #357, opened 2022-06-18
Warlock added and Javelin conditional set

TheGrayAlien:
Adds Warlock to the Turn Order after 1.16 update.
Conditions Javelin points only when Roots adventure is played.
(EX: My Ruleset 'Demeo Reloaded' has Javelins drop in any adventure)


jimconner (approved):
Awesome catch @TheGrayAlien - Thank you very much for submitting :-)

LGTM - Merging.

PR #362, opened 2022-06-19
Perform maintenance on RoomFinder.

orendain:
No function changes, except perhaps some very minor tweaks to element positioning.

**Changes:**
- Renamed `RoomListEntry` to `Room` for simplicity.
- Apply formatting (word-wrapping) to `Room.cs`.
- Remove redundant `this.` qualifiers.
- Remove `RoomListPanel.GameObject` to `RoomListPanel.Panel` for readability.
- Rename `SetRooms()` to `UpdateRooms()` for correctness.
- Separate the attach-object-to-parent logic from the creation of those objects.
- Remove unused object.
- Rename GameObject from "Rooms" to "RoomPage" for specificity.
- Move methods around to follow the method that calls them.
- Replace the parameter for `RoomListPanel.UpdateRooms()` from `RoomInfo` to `Room`.  Requiring the caller to handle the parsing from `RoomInfo` -> `Room` is meant to follow the single responsibility principle, and reduce nested method calling.
- Apply some formatting.
- Split `Render()` into `Draw()` and `RedrawRoomPages()`. This improves performance by only redrawing rooms on each refresh, and not static elements like sort buttons.
- Clear rooms from the stack when redrawing.
- Move header from `RoomFinderUI` to `RoomListPanel`, where it belongs better.
- Renamed `ModPatcher` to `Patcher` for both simplicity and correctness.  That is, the class performs patching, not patching on a mod.
- Renamed `ModState` to `SharedState`. Having global state like this, at all, is undesired, but does indeed seem to the be simplest approach for communicating between a static "Patcher" class and Unity `MonoBehaviour`s that don't take in arguments.

### DIFF
--- a/Common/UI/PageStack.cs
+++ b/Common/UI/PageStack.cs
@@ -35,7 +35,20 @@ namespace Common.UI
         public void AddPage(GameObject page)
         {
             _pages.Add(page);
+            UpdatePageStatus();
             UpdatePageVisibility();
+        }
+
+        /// <summary>
+        /// Removes all pages from the stack.
+        /// </summary>
+        /// <remarks>
+        /// Note that this stops tracking all current pages, but does not explicitly destroy them.
+        /// </remarks>
+        public void Clear()
+        {
+            _pages.Clear();
+            _currentPageIndex = 0;
             UpdatePageStatus();
         }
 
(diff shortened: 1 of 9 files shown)


jimconner (approved):
LGTM :-) (edit: whoops didn't notice this was still in draft status, but it still looks good to me)

PR #363, opened 2022-06-19
Add a `Clear` method for removing all pages from a page stack.

orendain:
Adds a `Clear` method for removing all pages from a page stack.

Without a way to clear pages, UIs with changing results (e.g., RoomFinder) only ever add paginated results and don't clear out old entries.


jimconner (approved):
You've certainly been busy.. Approved and merged :-)

PR #370, opened 2022-06-22
Refactor `common\ui` so as separate UI helpers based on UI environments.

orendain:
Refactor `common\ui` so as separate UI helpers based on UI environments.

⚠️ This PR **_breaks_** the `main` branch until [the subsequent PR](https://github.com/orendain/DemeoMods/pull/371) is merged in.


jimconner (approved):
The bits of this which I understand all look good, and the bits I don't understand... well... merging :-)

PR #379, opened 2022-06-22
Remove `ButtonMenuGreen` as a required dependency for NonVR UIs.

orendain:
Remove `ButtonMenuGreen` as a required dependency for NonVR UIs.  For compatibility with the 1.16 patch on 2022-06-22.

We aren't using this sprite.  Was added as a preliminary measure for future UI work.


jimconner (approved):
LGTM.

PR #380, opened 2022-06-22
Ignore amendment-labelled PRs when generating changelogs.

orendain:
Any PR labeled with `type: amendment` will be ignored by the changelog generator.


jimconner (approved):
Nice One!

PR #381, opened 2022-06-22
Introduce a RoomFinder UI for the PC Edition.

orendain:
Introduce a RoomFinder UI for the PC Edition.

**Changes:**
- Added NonVr UIs in the form of `RoomListPanelNonVr` and `RoomFinderUiNonVr`.
- Added a patch (in `Patcher.cs`) for the PC edition so that the main menu showed up after refreshing the list of available rooms. Else, the client is stuck with the loading screen.  No patch is needed for the VR version, as the VR version properly "goes back" to the main menu.
- Removed some missed debug print statements from previous commits.

### DIFF
--- a/Common/UI/PageStack.cs
+++ b/Common/UI/PageStack.cs
@@ -85,6 +85,11 @@ namespace Common.UI
 
         private void UpdatePageVisibility()
         {
+            if (_pages.Count < 1)
+            {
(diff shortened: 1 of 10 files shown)


jimconner (approved):
Awesome work once again @orendain

PR #384, opened 2022-06-23
Perform maintenance on Common module.

orendain:
Changes:
- Adjust field names. Remove unused fields.
- Aggregate environment-related logic to the common class `Environment.cs`.
- Rename common module's ModPatcher to Patcher.


jimconner (approved):
LTGM, and more importantly still builds OK for me.

PR #386, opened 2022-06-23
Apply maintenance to VR and NonVR UIs.

orendain:
**Changes:**
- Simplify unity object heirarchy for NonVR by way of reducing some nesting and using the built-in `transform` and `gameObject` in both `HouseRulesUiNonVr` and `RoomFinderUiNonVr`. This is in contrast to creating entirely new objects.
- Change the default header font size for NonVr header font. This allows NonVr UIs to call our convenient `CreateHeaderMenuText()` instead of configuring the text more granularly.
- Apply for formatting and space fixes.
- Prefer `Vector2` use over `Vector3` (which don't apply translations on the Z-axis anyway) for flatscreen UIs.
- Remove _background class fields when they could be contained as local variables.
- Rename RoomListPanel to RoomListPanelVr for consistency.
- Rename RulesetSelectionPanel* to RulesetListPanel* for consistency.


orendain:
Closing for https://github.com/orendain/DemeoMods/pull/387.

PR #388, opened 2022-06-23
Fix RoomFinder getting stuck on Oculus PC edition.

orendain:
**Problem:**
There was a race condition between when the game closes the "Loading..." screen and when the RoomFinder UI determines that the game's room-refreshing process is complete.  This caused two issues:
- The loading screen would never close, causing most of the UI to be unresponsive.
- Instead of closing the loading screen, the game would proceed to connect to the last-connected room.

**Fix:**
As part of the logic for determining if the game's room-refreshing cycle is complete, explicitly wait for the game to exit the loading screen.

Additionally, this PR renames `_roomListPanelVr` -> `_roomListPanel` and removes `_isInitialized` field.

### DIFF
--- a/RoomFinder/Patcher.cs
+++ b/RoomFinder/Patcher.cs
@@ -67,6 +67,7 @@
             }
 
             __instance.GetLobbyMenuController.view.ShowMainContent(LobbyMenuController.MenuContent.Play);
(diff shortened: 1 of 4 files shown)


jimconner (approved):
Nice work sir :-)

PR #390, opened 2022-06-23
Two new rules for preventing player on player accidental damage.

jimconner:
Before reviewing this PR:

Should I actually be raising this PR against `main`? This PR will be adding new rules to the README.md which won't work in the DLLs that we released yesterday. Maybe time to start using a `develop` branch?

# BlockFriendlyFireRule
Adds players to the list of targets who blocked an attack - damage is reduced to 1HP

This rule applies a prefix patch to `GenerateAttackDamage`. We already have a postfix patch on that method. Not sure if this might cause an issue?  

# DontShockFriendsRule
Zero all player on player electricial damage

This rule patches `DealDamage` (which was a challenge as it was an ambiguous match until I told it the param types) and will zero out player on player electrical damage only. This rule is probably a good candidate for the `Better Sorcerer` ruleset now that the Overcharge mechanic has been added to the game and zapping your friends is a common occurence.

### DIFF
--- a/HouseRules_Essentials/Rules/BlockFriendlyFireRule.cs
+++ b/HouseRules_Essentials/Rules/BlockFriendlyFireRule.cs
@@ -1,0 +1,81 @@
+﻿namespace HouseRules.Essentials.Rules
+{
+    using System;
+    using System.Collections.Generic;
+    using Boardgame;
+    using Boardgame.BoardEntities;
+    using Boardgame.BoardEntities.Abilities;
+    using Boardgame.Data;
+    using DataKeys;
+    using HarmonyLib;
+    using HouseRules.Types;
+    using UnityEngine;
+
+    public sealed class BlockFriendlyFireRule : Rule, IConfigWritable<bool>, IPatchable,
+        IMultiplayerSafe
+    {
+        public override string Description => "Friendly fire is blocked.";
+
+        private static bool _globalAdjustments;
+        private static bool _isActivated;
+
+        private readonly bool _adjustments;
+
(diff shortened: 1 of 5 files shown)


HouseRules_Essentials/Rules/BlockFriendlyFireRule.cs
```diff
@@ -0,0 +1,81 @@
...
+    using Boardgame;
+    using Boardgame.BoardEntities;
+    using Boardgame.BoardEntities.Abilities;
+    using Boardgame.Data;
+    using DataKeys;
+    using HarmonyLib;
+    using HouseRules.Types;
+    using UnityEngine;
+
+    public sealed class BlockFriendlyFireRule : Rule, IConfigWritable<bool>, IPatchable,
```
orendain, line 14: nit: Prefer [passive voice for rule ](https://github.com/orendain/DemeoMods/blob/701de544367af323754366293c10caa85f5a9b5e/HouseRules_Essentials/Rules/SampleRule.cs#L12-L15)naming.

orendain (commented):
I know it's preliminary, so just posting comments.  But does indeed LGTM.

> Maybe time to start using a develop branch?

Oh, right - great catch!  I'm up to try this.

> We already have a postfix patch on that method. Not sure if this might cause an issue?

Should be good!  From https://harmony.pardeike.net/articles/patching-prefix.html:

> The first prefix that returns false will skip all remaining prefixes unless they have no side effects (no return value, no ref arguments) and will skip the original too. **Postfixes and Finalizers are not affected.**

PR #392, opened 2022-06-29
Fix/Cleanup Demeo Reloaded ruleset

TheGrayAlien:
Fixed issue with shop floor not loading
Adjusted and cleaned up some invisible spawns
Added json to sample rulesets


jimconner (approved):
LGTM - Thanks @TheGrayAlien

PR #394, opened 2022-07-03
Fixes Sneak

TheGrayAlien:
Prevents Sneak from immediately replenishing after it fades.


jimconner (approved):
Nice One - Thanks @TheGrayAlien

PR #395, opened 2022-07-07
Use the assembly name when targetting the files to copy to the game directory.

orendain:
Technically, the assembly name is the correct value to use here.  It just so happens that the project and assembly share the same name.


jimconner (approved):
Nice catch.

PR #396, opened 2022-07-08
Fix for CardAdditionRule not working

TheGrayAlien:
Fixes field renamed in newest Demeo Version 1.19


orendain (approved):
Woot - awesome fix for 1.19 compatibility!

PR #398, opened 2022-07-08
v1.19_compatibility fixes

jimconner:
GetNewLevelSequence now requires an extra LevelSequence.Controltype to be passed as an argument.

There is now the choice of 'OneHero' or 'TwoHeroes' when generating a LevelSequence. I've pinned it to OneHero for now as that's what the rule was designed to work with.


TheGrayAlien (approved):
LGTM and works in all my testing. Great work as usual Jim!

PR #399, opened 2022-07-08
New rule for Electricity damage between party members.

jimconner:
This is the same code which was in the `DontShockFriends` rule which I previously did a WIP PR for, but with the rule renamed to match our naming standards.

* Incorporated into the BetterSorcerer ruleset, because it makes the sorcerer better.

To test:
Use better-sorcerer ruleset.
Start a skirmish game with Sorcerer + others. 
Use overcharge on the sorc straight away, and then descned into the dungon.
Use overcharge again on the sorc to cause Overload, zapping all the party members whilst they're still on the entrance steps. 

Expected Outcome:
No damage or stun should be applied to any party members, and 'No Damage' text should be displayed on the Host's screen.

### DIFF
--- a/HouseRules_Essentials/Rules/PartyElectricityDamageOverriddenRule.cs
+++ b/HouseRules_Essentials/Rules/PartyElectricityDamageOverriddenRule.cs
@@ -58,6 +58,7 @@
 
             if (attacker.IsPlayer() && targetPiece.IsPlayer() && damage.HasTag(DamageTag.Electricity))
             {
+                targetPiece.effectSink.SubtractHealth(0);
                 var localizedText = Traverse.Create(damage).Method("GetLocalizedText", paramTypes: new[] { typeof(string), typeof(bool) }, arguments: new object[] { "Ui/pieceUi/notification/damage/noDamage", false }).GetValue<string>();
                 Notification.ShowGoldenText(new Target(targetPiece).gameObject, localizedText);
 
--- a/HouseRules_Essentials/Rules/PartyElectricityDamageOverriddenRule.cs
+++ b/HouseRules_Essentials/Rules/PartyElectricityDamageOverriddenRule.cs
@@ -1,0 +1,104 @@
+﻿namespace HouseRules.Essentials.Rules
+{
+    using Boardgame;
+    using Boardgame.BoardEntities;
+    using Boardgame.BoardEntities.Abilities;
+    using Boardgame.GameplayEffects;
+    using Boardgame.SerializableEvents;
+    using DataKeys;
+    using HarmonyLib;
+    using HouseRules.Types;
+    using UnityEngine;
+    using System.IO;
+
+    public sealed class PartyElectricityDamageOverriddenRule : Rule, IConfigWritable<bool>, IPatchable,
+        IMultiplayerSafe
+    {
+        public override string Description => "Player on player electricity damage is zero.";
+
+        private static bool _globalAdjustments;
+        private static bool _isActivated;
+
+        private readonly bool _adjustments;
+        private static GameContext _gameContext;
+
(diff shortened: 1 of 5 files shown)


HouseRules_Essentials/Rules/PartyElectricityDamageOverriddenRule.cs
```diff
@@ -0,0 +1,104 @@
...
+            var attackerTarget = Traverse.Create(abilityContext).Field("attacker").GetValue<Target>();
+            var attacker = Traverse.Create(attackerTarget).Field("piece").GetValue<Piece>();
+
+            if (attacker.IsPlayer() && targetPiece.IsPlayer() && damage.HasTag(DamageTag.Electricity))
+            {
+                var localizedText = Traverse.Create(damage).Method("GetLocalizedText", paramTypes: new[] { typeof(string), typeof(bool) }, arguments: new object[] { "Ui/pieceUi/notification/damage/noDamage", false }).GetValue<string>();
+                Notification.ShowGoldenText(new Target(targetPiece).gameObject, localizedText);
+                _gameContext.serializableEventQueue.SendEventRequest(new SerializeableEventSetStatusEffects(targetPiece.networkID, new EffectStateType[] { }, new EffectStateType[] { EffectStateType.Stunned }));
+
+                return false; // Don't run the original OnStarted method.
```
orendain, line 76: When trying out this rule, my client was seeing damage while my host wasn't.

After some quick digging, I believe it due to all players invoking `ProjectileHitSequence#OnStarted`.  If the host patches/skips that method, it doesn't stop clients from doing so and applying the damage (exactly like the status effect that we're following up on and removing).

If we update a player's health, even by zero, the game will internally raise the `needsSync` flag for that particular stat (health).  This stat-auto-syncing is the reason why we don't have to force a board sync when hosts compute different damages.  Several ways to touch the stat, this is just the first I found: `targetPiece.effectSink.SubtractHealth(0);` 

(If only this kind of auto-syncing existed for everything else in the game, hehe).

orendain (approved):
Apologies for the late review. I renewed my fight with my computer's RAM and in turn was experiencing crashes when running the game for even a few minutes.  Though I'm now back to normal and able to test out this awesome new rule!

All LGTM with the exception that the client is still seeing the damage (and not being updated).

Approving now because I think the fix may be straight forward and I'd hate to block (more RAM changes tomorrow).

HouseRules_Essentials/Rules/PartyElectricityDamageOverriddenRule.cs
```diff
@@ -0,0 +1,104 @@
...
+                original: AccessTools.Method(typeof(ProjectileHitSequence), "OnStarted"),
+                prefix: new HarmonyMethod(
+                    typeof(PartyElectricityDamageOverriddenRule),
+                    nameof(ProjectileHitSequence_OnStarted_Prefix)));
+
+            harmony.Patch(
+                original: AccessTools.Method(typeof(SerializableEventQueue), "ValidateSerializableEvent"),
+                prefix: new HarmonyMethod(
+                    typeof(PartyElectricityDamageOverriddenRule),
+                    nameof(SerializableEventQueue_ValidateSerializableEvent_Postfix)));
```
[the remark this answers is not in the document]
jimconner: This patch shouldn't have been included at all... It was a WIP commit on the DontShockFriends branch from when I was trying to debug what was going on with the statusEffect sync. KHyperia suggested adding it as a workaround because the serializable event wasn't allowed by default. I've now removed it.

HouseRules_Essentials/Rules/PartyElectricityDamageOverriddenRule.cs
jimconner: Oh nice.. I've added that into the rule now and tested. Works a charm.

PR #400, opened 2022-07-08
Reposition and resize elements for a cleaner look.

jimconner:
Visual tweaks for NonVR HouseRules menu.

image

* Wider and slightly higher buttons to avoid text overlap.
* Reduced button text size from 25 to 22
* Applied RectTranform to buttonText object to enable wider text without line wrapping.
* Repositioned page elements for a more balanced layout.


orendain (approved):
Awesome new look!

PR #401, opened 2022-07-09
Fix Potion Stands loot

TheGrayAlien:
Prevents CardAdditionOverriddenRule from affecting Potion Stands.
Next logical step would be to add a method to generate custom random potion stand loot.


jimconner (approved):
Nice work @TheGrayAlien 👍

PR #402, opened 2022-07-09
Write a CI workflow to automate the building and assembling of mods.

orendain:
Write a CI workflow to automate the building and assembling of mods.


jimconner (approved):
Sterling work @orendain - Brilliant 👍

PR #405, opened 2022-07-10
Add workflow badge to README.

orendain:
This links to the list of all build runs since we [can't link _directly_ to the latest run](https://github.com/badges/shields/discussions/7247).


jimconner (approved):
👍

PR #408, opened 2022-07-12
Fix Desert Merchant Loot

TheGrayAlien:
Prevents CardAddition rule from affecting Jeweled Scarab turn-in to merchant
Might be a faster method than converting enum to int but I don't know it. Feel free to update!

### DIFF
--- a/HouseRules_Essentials/Rules/CardAdditionOverriddenRule.cs
+++ b/HouseRules_Essentials/Rules/CardAdditionOverriddenRule.cs
@@ -20,6 +20,7 @@
         private static Dictionary<BoardPieceId, List<AbilityKey>> _globalHeroCards;
         private static bool _isActivated;
         private static bool _isPotionStand;
+        private static int _numPlayers;
 
         private readonly Dictionary<BoardPieceId, List<AbilityKey>> _heroCards;
 
@@ -71,6 +72,7 @@
             Interactable whatIsit = gameContext.pieceAndTurnController.GetInteractableAtPosition(targetTile);
             if (whatIsit.type == Interactable.Type.PotionStand)
             {
+                _numPlayers = gameContext.pieceAndTurnController.GetNumberOfPlayerPieces();
                 _isPotionStand = true;
             }
         }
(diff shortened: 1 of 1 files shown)


HouseRules_Essentials/Rules/CardAdditionOverriddenRule.cs
```diff
@@ -88,15 +89,23 @@ private static void Patch(Harmony harmony)
...
             }
 
             var addCardToPieceEvent = (SerializableEventAddCardToPiece)request;
             var gameContext = Traverse.Create(__instance).Property<GameContext>("gameContext").Value;
 
             var pieceId = Traverse.Create(addCardToPieceEvent).Field<int>("pieceId").Value;
+            var cardSource = Traverse.Create(addCardToPieceEvent).Field<int>("cardSource").Value;
+
+            if (cardSource != (int)MotherTracker.Context.Energy && cardSource != (int)MotherTracker.Context.Chest)
+            {
```
orendain, line 105: In case you prefer using enums, I believe you may be able to replace `...<int>`:
```cs
Traverse.Create(addCardToPieceEvent).Field<MotherTracker.Context>("cardSource")
```

which would give you back something of type `MotherTracker.Context` instead of an int.  Either way is good though.

HouseRules_Essentials/Rules/CardAdditionOverriddenRule.cs
```diff
@@ -67,10 +67,11 @@ private static void Patch(Harmony harmony)
                 return;
             }
 
+            _isPotionStand = false;
             Interactable whatIsit = gameContext.pieceAndTurnController.GetInteractableAtPosition(targetTile);
             if (whatIsit.type == Interactable.Type.PotionStand)
             {
-                _isChest = false;
+                _isPotionStand = true;
```
orendain, line 74: QuickQuestion:  Unless `_isPotionStand` is reset when a card is granted, does this mean that this following scenario might happen?

- Player visits potion stand.  `_isPotionStand` set to `true`.
- Player gets an unaltered card.
- Player tops off their card energy.
- Since `_isPotionStand` is still `true`, player will get another unaltered card.
-

orendain (commented):
Awesome.  Just some clarifying questions.

HouseRules_Essentials/Rules/CardAdditionOverriddenRule.cs
TheGrayAlien: Didn't notice that, thanks! Should be fixed with my latest commit.

TheGrayAlien: Whoops, not ready yet. Forgot it will cycle through when multiplayer. Will fix that here in a few and update.

PR #410, opened 2022-07-23
Fix WaterBottleChest loot

TheGrayAlien:
Allow original loot for WaterBottleChest object on Desert campaign.


jimconner (approved):
Nice One :-)

PR #412, opened 2022-07-27
New Potion Commotion ruleset where all card additions are potions.

jimconner:
# ⚱️ Potion Commotion ⚱️

## What

A new ruleset idea suggested by a friend. Heroes start with their usual cards and abilities (except that Sorcerer's Zap is a 0AP casting cost `#STS`), but all of the cards that they get from chests or mana are Potions. Enemy respawns are disabled.

We gave this a playtest on Elven Necropolis thie evening and very much enjoyed it. This ruleset would be a real struggle on rat king as you'd only have the torches from the shop (if you're lucky enough to get them).

## How to review

Build from this branch and give it a playtest.


TheGrayAlien:
I'll give it a build and try it out tonight with a few others.

TheGrayAlien:
Played it with friends and have a few notes.

- More chests and/or faster mana regen to keep the potions flowing

- Decrease the chance of getting the one-time-use damage resist potion by doubling or tripling the number of other potions in the CardAddition list. We were all resist maxed out and just recycling those halfway through floor 2.

- Why is (only) Zap free? Were we all supposed to be sorcerers? (3/4 of us were, but just asking)

It's playable as-is and we won in the end but it seemed a bit slow moving. By the end of level 2 we almost all were card-starved and not quite max stats. I know we can't change multiplayer shops so everyone bought non-potions while we were there. If that's not intended then maybe less gold per floor would reduce shop reliance.

jimconner:
Which map did you try the ruleset on @TheGrayAlien ? We were desperate for more DamageResistPotions when we played it, but we had been putting them on Cána. We were just about maxxed out by the time we made it to the boss fight (which was a walk in the park). Our card starvation happened near the end of the 1st floor in our game.

Zap is 0AP becuase without that the Sorcerer is damage-limited compared to other heroes - As Sorcerer can no longer get Fireball/Freeze cards removing the casting cost on Zap evens things up a bit.

TheGrayAlien:
> Which map did you try the ruleset on @TheGrayAlien ? 
We just ran Black Sarcophagus

> Zap is 0AP because without that the Sorcerer is damage-limited compared to other heroes - As Sorcerer can no longer get Fireball/Freeze cards removing the casting cost on Zap evens things up a bit.
I see. Might consider making Overcharge and/or Lightning Bolt also free for the same reason since none of them can be used in the same turn.

TheGrayAlien (approved):
Let's do it!

PR #416, opened 2022-09-12
Make Roomcode compatible with v1.20 of Demeo.

jimconner:
CreatingGameState is no longer part of GameStateMachine, but has instead been moved into it's own class.

Updated patcher to reflect this change. Tested it on the PC version and it got my user-defined roomcode OK.


TheGrayAlien (approved):
LGTM!

PR #417, opened 2022-09-12
Ensure compatability with v1.20

jimconner:
* Also work around a melonloader compatibility issue with the new version of Unity being used by the latest patch.


TheGrayAlien (approved):
LGTM!

PR #422, opened 2022-10-08
Fix exploit to run public games

TheGrayAlien:
Using 'Play Again' as host would bypass the private room check and activate code in public games


jimconner (approved):
Nice catch @TheGrayAlien  :-)

PR #425, opened 2022-10-25
Passthrough version fix

TheGrayAlien:
Builds fine and seems to work.
Might not be the best way to handle the lack of PieceType.Enemy now, though.


jimconner (approved):
Nice work @TheGrayAlien . LGTM

PR #429, opened 2022-10-28
Fix bug with focusing on marked targets

TheGrayAlien:
OOPS! Pets weren't attacking marked targets... enemies were!


jimconner (approved):
LGTM :-)

PR #431, opened 2022-11-07
Fix Ability Cooldowns

TheGrayAlien:
Fix abilities with cooldowns set greater than 1 turn
(Note: this works to fix the problem but the underlying cause of why cooldowns are being set to -1 should be looked into)


jimconner (commented):
Awesome work @TheGrayAlien - Thank you for fixing.

jimconner (approved):

PR #433, opened 2022-11-22
Fix health and Attack scaling

TheGrayAlien:
And also Enemy door opening rule.


TheGrayAlien:
Due to execution of this rule on every instance of each piece it was increasing health/attack exponentially for each creature of the same type spawned. The method I used to prevent this might not be the proper one, but it works. If there is a better method please let me know! Thanks.

jimconner (approved):
This looks awesome to me. Nice going @TheGrayAlien

PR #434, opened 2022-11-28
Fixes for movement issues that need sync'd properly

TheGrayAlien:
Fixes Darkness/Fog of War not always lifting when moving
Fixes disappearing characters (players and enemies) after movements
Fixes warping back when getting grabbed/pulled by enemies


jimconner (approved):
Amazing work :-)

PR #439, opened 2022-12-02
Fixes for 12/1 update

TheGrayAlien:
Also FogAndSpawn change in Boardsyncer.


jimconner (approved):
Nice work @TheGrayAlien

PR #440, opened 2022-12-08
Fix missing spawns and MonsterDeck

TheGrayAlien:
Fixes issue with lack of enemy spawns since last update
Fixed MonsterDeckOverridden rule to allow the Boss and its' spawns to work again.


jimconner (approved):
Nice work @TheGrayAlien

PR #442, opened 2022-12-15
V1.25 fixes

jimconner:
# Demeo v1.25 fixes
Here are some fixes for making HouseRules compile against v1.25. They're largely untested aside from making sure that they compile and that I can launch a game with a few different rulesets. Actually playing a game has not been attempted with these 'fixes'

## Changes:
* `Interactable` now has an added override method for OnInteractable which was causing `Harmony.patch` to complain about an Ambiguous method being used in EnemyHealthScaledRule - The fix for this was to specify an array of the type parameters which are being accepted by the method we want to patch.
* `Inventory.flags` has been changed from an `int` to an `enum` meaning that CardAdditionOverriddenRule was complaing about being unable to use boolean operations on an enum. I attmpetd to change the code to maniuplate the enum directly, but the numeric values that I'm using are still from the old boolean code - I think this is OK, but we really need to check that card replenishes are working correctly.
* The game has change from only having a single `weakened` state to having two different ones for 1 & 2 turn duration respectively. I updated the rulesets where we were giving immunity to weakened to apply both immunities.
* `PieceConfigData` no longer has a `PieceName` field. It does however have a `PieceNameLocalizationKey` which is hopefully the same thing - I updated the code accordingly and nothing crashed when I tested launching a few different rulesets.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Prepare
         run: |
-          C:\msys64\usr\bin\wget.exe https://gist.github.com/jimconner/f3c89c300d0b13e23c32fc9fb5d1b1df/raw/a18488125ab3d34aaca7b39350d4e6119b6371fc/ws.zip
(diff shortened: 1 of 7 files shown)


TheGrayAlien:
LGTM! I'll be adding a few fixes in rules and rulesets specifically for the Barbarian shortly.

TheGrayAlien (approved):

PR #443, opened 2022-12-16
Add the Barbarian

TheGrayAlien:
Barbarian added to rules/rulesets and minor ruleset updates


jimconner (approved):
Awesome work as always @TheGrayAlien

PR #444, opened 2022-12-16
More ruleset updates

TheGrayAlien:
Ruleset fixes due to me assuming some incorrect Barbarian ability names
Switch health/attack scale method away from name localization changes
'Flipping Out!' ruleset was missing from the solution file


jimconner (approved):
Nice work @TheGrayAlien

PR #445, opened 2022-12-16
Missed a closing brace

TheGrayAlien:
Oops :P


jimconner (approved):
LGTM, but there is no need to be updating these rulesets manually - If you uncomment the appropriate line in HouseRules_Configuration/ExampleRulesetExporter.cs then it will automatically export all of the rulesets into UserData/HouseRules when the game starts up.

PR #446, opened 2022-12-17
Fix CardAddition bug

TheGrayAlien:
Fix wrong variable so it was never reset after opening Vortex Dust chest resulting in normal cards
instead of user defined cards the rest of the game


jimconner (approved):
Nice catch.

jimconner (approved):

PR #457, opened 2023-01-04
Fix Boss specific maps

TheGrayAlien:
Serpent Lord and Elven King require specific maps for 3rd floor or game will hang forever at map loading


jimconner (approved):
Nicely done Sir :-)

PR #458, opened 2023-01-20
Initial upload of new mod

TheGrayAlien:
Introduction to a new mod to view text with advanced stats of each player piece

### DIFF
--- a/AdvancedStats/AdvancedStats.csproj
+++ b/AdvancedStats/AdvancedStats.csproj
@@ -68,7 +68,7 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <Target Name="CopyToModsDir" AfterTargets="Build">
-    <Copy SourceFiles="$(OutputPath)\$(ProjectName).dll" DestinationFolder="$($(DemeoPcDir))\Mods" />
+    <Copy SourceFiles="$(OutputPath)\$(ProjectName).dll" DestinationFolder="$(DemeoVrDir)\Mods" />
     <Copy SourceFiles="$(OutputPath)\$(ProjectName).dll" DestinationFolder="$(DemeoPcDir)\Mods" />
     <Copy SourceFiles="$(OutputPath)\$(ProjectName).dll" DestinationFolder="$(DemeoOculusVrDir)\Mods" />
     <Copy SourceFiles="$(OutputPath)\$(ProjectName).dll" DestinationFolder="$(DemeoOculusPcDir)\Mods" />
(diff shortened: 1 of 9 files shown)


AdvancedStats/AdvancedStats.csproj
```diff
@@ -0,0 +1,76 @@
...
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="0Harmony">
+      <HintPath>..\..\..\..\..\..\Program Files (x86)\Steam\steamapps\common\Demeo\MelonLoader\0Harmony.dll</HintPath>
```
jimconner, line 37: These hardcoded HintPaths don't work for me. I get lots of errors like the one below because it can't find the required references at the HintPath defined in the .csproj.

```
Error	CS0246	The type or namespace name 'HarmonyLib' could not be found (are you missing a using directive or an assembly reference?)
```

It would be better to use the same $(DemeoVrDir) for this as we do with the other modules. We're already telling people that they need to set this ENV var in our docs, so makes sense to keep on using it..

jimconner (changes requested):

AdvancedStats/AdvancedStats.csproj
jimconner: I pushed another commit to fix the reference paths. Just waiting for the checks to complete and then I'll merge.

jimconner (commented):
I'm having issues getitng this to complile. Inside NonVRAdvancedStatsView there are references to `NonVRInfoPanelController` and `PiceDetectionData` which are unresolvable. I've tried searching for these things with dnspy and ack/grep, but can't find them mentioned in the game code. Any idea what I'm missing here?

jimconner:
This doesn't appear to be getting built as part of our github workflows currently - we should probably be including it in .github/workflows/build.yml so that the github tests remain valid.

TheGrayAlien:
> I'm having issues getitng this to complile. Inside NonVRAdvancedStatsView there are references to `NonVRInfoPanelController` and `PiceDetectionData` which are unresolvable. I've tried searching for these things with dnspy and ack/grep, but can't find them mentioned in the game code. Any idea what I'm missing here?

Yes, this code only exists in the Desktop version of the CSharp DLL. It was tricky for me to discover it due to this and it's why my class checks to see if the game is running on Desktop before trying to execute it. If you switch your build to the PC Edition (Steam OR Oculus) you will see that code.

TheGrayAlien:
> This doesn't appear to be getting built as part of our github workflows currently - we should probably be including it in .github/workflows/build.yml so that the github tests remain valid.

I'm assuming you know how to do that because I sure don't (yet!)

jimconner (approved):
Nice work @TheGrayAlien 🥇

PR #460, opened 2023-02-22
Fixes for PS5 update

TheGrayAlien:
Fixes House Rules
Fixes RoomFinder
Fixes RoomCode

They just renamed a variable and changed the location of a function...

Note: This isn't currently building because our binaries for the current Demeo build aren't on the server.

Example (D:\a\DemeoMods\DemeoMods\RoomFinder\RoomFinderMod.cs(26,39): error CS0117: 'MotherbrainGlobalVars' does not contain a definition for 'IsRunningOnNonVRPlatform')

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Prepare
         run: |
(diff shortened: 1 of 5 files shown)


jimconner (approved):
Nice work @TheGrayAlien - Sorry it took me so long to update the bundle.

PR #461, opened 2023-03-15
New Rule for Potion Stands

TheGrayAlien:
Lets user configure potential card list for Potion Stands


jimconner (approved):
Fantastic work @TheGrayAlien - Thank you for keeping the dream alive :-)

PR #463, opened 2023-03-22
Two new rules added to specify Energy cards gained and/or Chest cards gained

TheGrayAlien:
CardAddition will still work so no need to update existing rulesets
Standardized how randomness is computed between all loot sources
Fixed bug in PotionAddition and CardAddition that ignored the first two loot items
Removed leftover log from StartCardsModified
Updated README to reflect changes/additions


jimconner (approved):
Excellent work @TheGrayAlien 💯

PR #464, opened 2023-03-22
New Rule ApplyEffectOnHitAdjusted

TheGrayAlien:
Adjusts the effect that a BoardPiece has on attackers


jimconner (approved):
I'm running out of new ways to say 'Excellent Work' - Nice one @TheGrayAlien

PR #466, opened 2023-04-27
Update Build Version

TheGrayAlien:
For consistency and to prevent "newer version available" message.


jimconner (approved):
Good catch @TheGrayAlien

PR #468, opened 2023-05-17
Separate GitHub workflow CIs into mod-specific individual ones.

orendain:
Breaks up the single GitHub workflow into ones that are mod-specific.

Additionally:
- Updates some of the GitHub Actions (e.g., `setup-msbuild@1.1` -> `setup-msbuild@1.3.1`).
- Add a path trigger to the corresponding mod's subdirectory.  CIs will only run if its mods' files are modified.

If approved, we will also need to update [the branch status check rules](https://github.com/orendain/DemeoMods/settings/branch_protection_rules/24009555), so that previous CI steps are no longer required to pass (note at the bottom of this PR that GitHub is saying some checks have not completed yet), while new ones are.  That can be done (by any repo contributor, AFAIK) before merging, if this PR change is acceptable.


jimconner (approved):
Nice one @Mex

PR #470, opened 2023-05-17
Update HouseRules for compatibility with Demeo's 2023-05-16 update.

orendain:
> **Note**
> 
> WIP. Dumping the changes, and notes, in the meantime for reference.

So far with this PR, HouseRules seems to be working, though I tested only a single game and only for half a map.

Current Changes:
- Seems that the scene index for hangouts has changed from `43` to `45.  Updated `Common/UI/Environments.cs` to treat scene `45` as representing Hangouts being loaded.
- Large chunks of the code dealing with the game table in Hangouts has been moved around.  For example, HouseRules' `HangoutsGameRacer.cs` would look to code in the game's `GroupLaunchTable.cs`.  The new update has no such file, but rather has a similar (but modified) version as `GroupLaunch.cs`.  This new `GroupLaunch.cs` is not included in the standard assembly, but rather packaged separately as part of `GroupLaunchPartyRuntime.dll`.
   - [Originally writing `HangoutsGameRacer`](https://github.com/orendain/DemeoMods/pull/267) was a challenge to begin with because of the how many steps are involved inbetween the time a game is started in Hangouts and when the _actual_ room is launched.  Rather than attempting to reconstruct that effort with the game's modified Hangouts code, this PR removes `HangoutsGameRacer`.  The removal need not be permanent - it was done so as to be able to fix HouseRules without an additional class to consider.
  - Without `HangoutsGameRacer`, games started from Hangouts will _likely_ still work ... though potentially not as often as they would otherwise as other players may beat the local player into a room, gaining them the host spot.
- Buttons in Hangouts were not responding to the cursor.  Turns out that buttons in Hangouts now need more components in order to register as "interactible".
  - Note: I will whip up a summary (as part of this PR) of how I typically debug issues like this, and how I figured out what had to be done to fix our buttons working in Hangouts.
  - TODO: Will need to circle back later and clean up some bits that are no longer required.  For example, note that a call to `CommonModule.HangoutsButtonHandler.RegisterBowserButton()` is commented out as the way to register selectable buttons has changed.
- Up until now, "Page" (i.e., the pre-fabricated windows used for our mod menus) navigation was rendering as typical VR buttons, as opposed to VR + special-Hangouts-tweaks buttons.  Modified `Common/UI/PageStackNavigation.cs` to use the proper button creator when pages are rendered in Hangouts environments.
  - Note: Will attempt to extract this change into a PR of its own for simplicity.
- `Notification.ShowGoldenText()` has changed.  Instead of trying to fix this properly, this PR temporarily commented it out for dev purposes.


orendain:
Not finished, but figured I'd dump a status update and share a preliminary working (mostly?) build.  CC @TheGrayAlien @jimconner .

TheGrayAlien:
This now conflicts because I had already approved the RoomFinder PR.

Also, this wouldn't work due to overloaded function error until I changed `Common\UI\` line 21 of `PageStackNavigation.cs` to `public static PageStackNavigation NewInstance(PageStack pageStack, IElementCreator elementCreator)` and it compiles, but I'm not even certain that was the correct way to handle that.

orendain:
Thanks a lot @TheGrayAlien for the findings in https://github.com/orendain/DemeoMods/pull/471#issuecomment-1551607686.  I found the issues in the Oculus version and fixed them in this PR (added more findings in the main PR post).

I think this PR is ready for review, if the `TODO`s listed in the main post above are acceptable to do post-merge in other PRs.

TheGrayAlien (approved):
Great work! So glad you've come back to help with this!

PR #471, opened 2023-05-17
Dependency inject an `IElementCreator` into Page stacks rather than creating one internally.

orendain:
Discovered while working on https://github.com/orendain/DemeoMods/pull/470.

This PR aims to remove logic from `Common/UI/PageStackNavigation.cs` so that instead of trying to determine which kind of environment (PC, VR, Hangouts) is currently loaded so that an appropriate ElementCreator (for creating buttons/text) is used, it instead expects an ElementCreator to be passed in directly in standard dependency injection form (i.e., in the constructor).

Since the correct `ElementCreator` is already known at a higher level (e.g., `RoomFinder/UI/RoomListPanelNonVr.cs` and `RoomFinder/UI/RoomListPanelVr.cs`), giving `PageStackNavigation.cs` what it needs is just a matter of passing down values that are already known, and not rely on it being looked-up internally.

Moving this it's own PR to make https://github.com/orendain/DemeoMods/pull/470 that much less large.

---

Some more info just for context and knowledge share:

We have created 3 "element creators" to help facilitate easy text/button rendering:
- `Common/UI/Element/NonVrElementCreator.cs`
- `Common/UI/Element/VrElementCreator.cs`
- `Common/UI/Element/HangoutsElementCreator.cs`

It turns out that `Common/UI/PageStackNavigation.cs` has only been using the VR and Non-VR varients, and not the Hangouts variant (note how the code this PR removes does not reference Hangouts).  While this seems to have worked thus far, this could be considered a bug.  Up until now, buttons in "pages" (i.e., the pre-fabricated windows that hold our mods' menus) rendered in Hangouts have been treated as non-Hangouts VR buttons, and have been missing the special Hangouts "tweaks" that allow them to be compatible with Hangouts as much as possible.

This PR now expects the correct element creator to be passed in, guaranteeing that menu pages in Hangouts use the hangouts element creators passed in from the top.


TheGrayAlien:
Tested... This works great for main and Hangouts UI's on Steam VR and PC, and Oculus PC but NOT for Oculus VR for some reason. I still have not button interaction in HouseRules and RoomFinder there for some reason..?
Note: Whenever I tried to enter Heroes Hangout I got a black screen bug of some kind so couldn't test inside there.

orendain:
> Tested... This works great for the Main HouseRules menu and Hangouts UI's on Steam VR. Also works great for PC and Oculus PC but NOT for Oculus VR for some reason. I still have no button interaction in HouseRules and RoomFinder there for some reason..?
Note: Whenever I tried to enter Heroes Hangout in Oculus VR I got a black screen bug of some kind so couldn't test inside there.
Update: I finally got into Heroes Hangout and there was no HouseRules menu to be found! Weird.

This PR by itself doesn't do much - it's just pulling out a single, focused, issue from https://github.com/orendain/DemeoMods/pull/470 so that https://github.com/orendain/DemeoMods/pull/470 isn't so large.

This PR on it's own just cleans up a process, but the proper fix for getting HouseRules fully working again would be part of https://github.com/orendain/DemeoMods/pull/470 (which would build on top of this PR once merged).  Hope that makes sense, and apologies for the misunderstanding!

TheGrayAlien (approved):
Works as intended, great work!

PR #472, opened 2023-05-19
Fixes Oculus HouseRules menu issues

TheGrayAlien:
Fix not being able to interact with Oculus main menu
Fix Oculus menu not appearing inside Heroes Hangout

Note: Because of the recent patch there are now different data files per platform. This means we're now having to look in a Steam game location for everything else and an Oculus game location for this one file. Also, for this to work you would need to have installed Demeo in the default directory given by the Oculus Windows app.


TheGrayAlien:
Ignore this :P Apparently my version was corrupted because the menus are working on the main build without this 'fix' now.

PR #473, opened 2023-05-19
Update Build Version for new release

TheGrayAlien:
Build version updated to 1.6.2 so we can release!


orendain (approved):
LGTM!  I would even say that a minor version bump to `1.x.0` is in order given how much has changed.  Either way, looks excellent.

PR #474, opened 2023-05-19
Apply general maintenance.

orendain:
Hopefully the size of this PR is not too overwhelming to review.  Can split up the PR is it is.  There is no change in functionality in this PR, it is simply a maintenance and cleanup pass.

Changes:
- Replace deprecated/obsolete MelonLoader methods with preferred ones (e.g., `OnApplicationStart` -> `OnInitializeMelon`).
- Use MelonLoader-provided Harmony instance instead of creating new instacnes of them.
- Fix formatting.
- Remove unused imports.
- Use `var` style declaration instead of explicit types.
- Remove redundant code.
- Fixed some capitilization.
- Reduced nesting by restructuring some condition flows.  Specifically, by [using guard clauses](https://refactoring.guru/replace-nested-conditional-with-guard-clauses).
- Update MelonLoader dependencies in the GitHub workflows.

### DIFF
--- a/HouseRules_Essentials/Rules/StartCardsModifiedRule.cs
+++ b/HouseRules_Essentials/Rules/StartCardsModifiedRule.cs
@@ -62,60 +62,66 @@
             }
 
             __result = false;
-            for (int i = 0; i < piece.inventory.Items.Count; i++)
+            for (var i = 0; i < piece.inventory.Items.Count; i++)
             {
-                Inventory.Item value = piece.inventory.Items[i];
+                var value = piece.inventory.Items[i];
 
-                if (value.IsReplenishing)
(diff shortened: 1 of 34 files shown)


HouseRules_Essentials/Rules/StartCardsModifiedRule.cs
orendain, line 1: FYI that the way GitHub is rendering this change makes it seem like the entire flow was rewritten, but it was not.  I inverted a couple of the `if` statements in order to reduce nesting and make the flow easier to follow, with no change to functionality, but GitHub isn't able to render that in a way that is clear what happened.

TheGrayAlien (approved):
LGTM, all seems compile and game seems to run fine

PR #477, opened 2023-05-25
Fixes menu issues...

TheGrayAlien:
Newer STEAM Demeo downloads only have 1 folder for both PC and VR
Fixes a bug with HouseRules and RoomFinder menus not appearing for that combined version


TheGrayAlien:
Tested and as this only affects when the menu is displayed it really can't (and seemingly doesn't) cause any issues checking for both possible Lobby Scene Indexes.

PR #478, opened 2023-06-15
New rule and ruleset using it

TheGrayAlien:
For those who need a new challenge... Hardcore is now a thing.


jimconner (approved):
What a brilliant new rule - nice one @TheGrayAlien

PR #479, opened 2023-07-05
In game rules list UI

TheGrayAlien:
VirtualDesktop Android-20230705-113604
VirtualDesktop Android-20230705-113633
VirtualDesktop Android-20230705-113703

(VR only) Creates a persistent panel next to the gaming table (only while playing) displaying the ruleset, description, and the rules used.

Changed lots of rule descriptions to be easier for non-programmers to understand. Also colorizes, shortens, and lowers the timer on the ShowCameraMessage. While I was there I removed some unnecessary code in the PieceDownedCount rule.
 
This would be **EASILY** modified to display user-configured rules descriptions if someone could create a way for me to get those... I am totally lost in how to do that.

### DIFF
--- a/HouseRules_Essentials/Rules/AbilityDamageOverriddenRule.cs
+++ b/HouseRules_Essentials/Rules/AbilityDamageOverriddenRule.cs
@@ -9,7 +9,7 @@
 
     public sealed class AbilityDamageOverriddenRule : Rule, IConfigWritable<Dictionary<AbilityKey, List<int>>>, IMultiplayerSafe
     {
-        public override string Description => "Ability damage values are overridden";
+        public override string Description => "Ability damage is adjusted";
 
         private readonly Dictionary<AbilityKey, List<int>> _adjustments;
         private Dictionary<AbilityKey, List<int>> _originals;
--- a/HouseRules_Essentials/Rules/BackstabConfigOverriddenRule.cs
+++ b/HouseRules_Essentials/Rules/BackstabConfigOverriddenRule.cs
@@ -10,7 +10,7 @@
     public sealed class BackstabConfigOverriddenRule : Rule, IConfigWritable<List<BoardPieceId>>, IPatchable,
         IMultiplayerSafe
     {
-        public override string Description => "Backstab is allocated to new BoardPieceIDs";
+        public override string Description => "Hero ability to backstab is adjusted";
 
         private static List<BoardPieceId> _globalAdjustments;
         private static bool _isActivated;
--- a/HouseRules_Essentials/Rules/FreeAbilityOnCritRule.cs
+++ b/HouseRules_Essentials/Rules/FreeAbilityOnCritRule.cs
@@ -11,7 +11,7 @@
     public sealed class FreeAbilityOnCritRule : Rule, IConfigWritable<Dictionary<BoardPieceId, AbilityKey>>, IPatchable,
         IMultiplayerSafe
     {
-        public override string Description => "Critical Hit gives free card.";
+        public override string Description => "Hero critical hits gives a free card";
 
         private static Dictionary<BoardPieceId, AbilityKey> _globalAdjustments;
         private static bool _isActivated;
--- a/Common/UI/Element/HangoutsElementCreator.cs
+++ b/Common/UI/Element/HangoutsElementCreator.cs
@@ -1,6 +1,7 @@
 ﻿namespace Common.UI.Element
 {
     using System;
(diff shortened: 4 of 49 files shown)


HouseRules_Essentials/Rules/AbilityDamageOverriddenRule.cs
```diff
@@ -9,7 +9,7 @@
 
     public sealed class AbilityDamageOverriddenRule : Rule, IConfigWritable<Dictionary<AbilityKey, List<int>>>, IMultiplayerSafe
     {
-        public override string Description => "Ability damage values are overridden";
+        public override string Description => "Ability damage is adjusted";
```
jimconner, line 12: In the past Mex has preferred the rule descriptions to match the rule names - An 'adjust' rule is adding a +_ modifier to an existing value, whereas an override rule sets a value directly. I don't have strong feelings on this myself, but thought I'd mention as I've been asked to change from 'adjusted' to 'overridden' in the descriptions before.

HouseRules_Essentials/Rules/BackstabConfigOverriddenRule.cs
```diff
@@ -10,7 +10,7 @@
     public sealed class BackstabConfigOverriddenRule : Rule, IConfigWritable<List<BoardPieceId>>, IPatchable,
         IMultiplayerSafe
     {
-        public override string Description => "Backstab is allocated to new BoardPieceIDs";
+        public override string Description => "Hero ability to backstab is adjusted";
```
jimconner, line 13: Non-hero characters can also be granted backstab with this rule.

HouseRules_Essentials/Rules/FreeAbilityOnCritRule.cs
```diff
@@ -11,7 +11,7 @@
     public sealed class FreeAbilityOnCritRule : Rule, IConfigWritable<Dictionary<BoardPieceId, AbilityKey>>, IPatchable,
         IMultiplayerSafe
     {
-        public override string Description => "Critical Hit gives free card.";
+        public override string Description => "Hero critical hits gives a free card";
```
jimconner, line 14: Nit picking... Too many plurals... 'hits gives' should be 'hit gives'

jimconner (approved):
Looks like amazing work. I'm marking this as approved and will leave the clicking of 'Merge' to you in case you wish to make any amendments. Awesome work @TheGrayAlien 🥇

TheGrayAlien:
> Looks like amazing work. I'm marking this as approved and will leave the clicking of 'Merge' to you in case you wish to make any amendments. Awesome work @TheGrayAlien 🥇

Thanks. If you have any insights on how to add personalized rule descriptions within rulesets that would help. I'd love the descriptions to be less vague and more specific according to the ruleset creators.

jimconner:
We already have in each of our rulesets a 'const string description' which we're using to store a single-line ruleset description. Couldn't we just add a `const string long_desc` and stick a couple of paragraphs of text into there?

TheGrayAlien:
I see what you're saying but what I would prefer is descriptions for each rule on the Ruleset level inside the JSON, not in the code of the rules. That would allow ruleset creators to define what is shown about what they did with each change. Make it optional, if possible, so everyone's current JSON ruleset files don't require changes. Is that something that can be done?

jimconner:
The json directly transcribes into the same C# values, so JSON ruleset creators would be able to add a long_desc field too. They would of course have to escape newline chars as JSON doesn't do multi-line strings. If a JSON ruleset long_desc is defined, we display that, otherwise we display the rule-based descriptions instead.

TheGrayAlien:
Interesting, I had no idea... it seems too easy :) I'll have a go at it next.

jimconner:
I'm not at all sure, but we might have to find a way of setting a default value if the field is undefined in the JSON when it tries to unpack it into the C# struct - I know that would be an issue in GoLang - can't remember if the C# json import has a similar restriction... Try it and see I guess.

jimconner:
Actually, that bit looks easy to overcome... We can just set a default if not-found before we return the ruleset.

https://github.com/orendain/DemeoMods/blob/main/HouseRules_Configuration/ConfigManager.cs#L145

TheGrayAlien:
Seems to be working as intended. Just a few tweaks and I'll send a PR. 
VirtualDesktop Android-20230705-170535
VirtualDesktop Android-20230705-170614
VirtualDesktop Android-20230705-170647

PR #480, opened 2023-07-05
Long descriptions for rulesets

TheGrayAlien:
Allow ruleset creators to describe their changes in detail instead of presenting a list of rules that were changed.
This doesn't require existing JSON's to be changed in order to work!
VirtualDesktop Android-20230705-170535
VirtualDesktop Android-20230705-170614
VirtualDesktop Android-20230705-170647


jimconner (approved):
Brilliant !!!

PR #482, opened 2023-07-15
Fix mis-spelled rule

TheGrayAlien:
StatModifiersOverridden was spelled wrong as StatModifiersOverriden
Thanks to @BobtheBunny from our Discord channel for bringing this to light.


jimconner (approved):
Good catch - Thank you @TheGrayAlien

PR #484, opened 2023-07-21
Fix blank Longdesc

TheGrayAlien:
Display rules instead of creator's description if the ruleset creator added Longdesc but used a blank placeholder of "".


orendain (approved):
LGTM.  Optionally, `string.IsNullOrEmpty()` is a helper that does both of these checks, in case you find that cleaner.

PR #485, opened 2023-07-27
Update README.md

TheGrayAlien:
Fixed display error not ending json code before next example


jimconner (approved):
Nice catch - Thanks @TheGrayAlien

PR #487, opened 2023-09-07
Fix HouseRules for 9/7/23 Demeo update

TheGrayAlien:
RoomFinder is broken (again) - Will look at that later
There's also a new MelonLoader error logged but HouseRules seems to work fine in all my tests


jimconner (approved):
Excellent work as always @TheGrayAlien . Thanks.

jimconner:
Looks as if the build dependencies bundle needs updating again as thje HouseRules_Essentials build can't find the new method names.

PR #488, opened 2023-09-08
Update RoomFinder to be compatible with Demeo 2023/09/07 update.

orendain:
Demeo changed the `lobby` field type of `gameStateMachine` from `Lobby` to `ILobbyMenuPresenter`, so fields/methods that were previously available no longer are.  This PR fetches necessary fields through another path.


TheGrayAlien (approved):
Great to see you back again @Mex! Always at a time of need... like Batman.

Come to think of it, I've never seen you and Batman at the same time...

PR #491, opened 2023-09-08
Update build dependencies for 2023-09-08.

orendain:
This is the HouseRules-equivalent of https://github.com/orendain/DemeoMods/pull/489.

PR #492, opened 2023-09-08
Update the `author` field to correctly reflect the team behind the develpment effort.

orendain:
Originally, the `author` field was set so as to attribute the mods to their corresponding GitHub repository.  However, in practice, that field is used more for display purposes.  Thus, this PR updates the `author` field so that it correctly reflects the folks behind the effort.

Should pass all build checks once https://github.com/orendain/DemeoMods/pull/491 is merged.


TheGrayAlien (commented):
Not yet passing the build check

TheGrayAlien (approved):

PR #495, opened 2023-09-18
Introduce Highlighter Mod.

orendain:
This new mod, Highlighter, provides in-game highlighting/coloring.  Currently, the only highlighting this mod makes available is for displaying a character's line-of-sight.

This was inspired by [a mod written by MuKen](https://steamcommunity.com/app/1484280/discussions/0/3091137796295708382/) which was once in the scene.

Features:
- When hovering the local player's character over a square, the board will be highlighted with all squares that would be in that character's line-of-sight if they were to move there.

Known Functional Concerns:
1. There is a limit to how far the line-of-sight highlighting reaches.  This is due to the built-in ability, `MagicMissle`, that the mod uses to compute which squares are reachable.  This can be improved if a better, further-reaching ability is discovered, or if we tweak how valid squares are computed.

[Screencast from 09-18-2023 02:28:54 AM.webm](https://github.com/orendain/DemeoMods/assets/1426304/0d7383e6-5e62-4a21-aae2-584487e28457)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,6 +5,10 @@
 'area: common':
 - Common/**
 
+'area: highlighter':
+  - .github/workflows/highlighter.yml
+  - Highlighter/**
+
 'area: houserules/configuration':
 - .github/workflows/houserules.yml
 - HouseRules_Configuration/**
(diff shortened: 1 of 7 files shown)


TheGrayAlien (approved):
Players will love this, I'm sure!

PR #497, opened 2023-09-19
Adust file naming to be consistent with how files are referenced.

orendain:
Otherwise, some IDEs complain about files (w/ case sensitivity) not being found.


TheGrayAlien (approved):
Nice

Issue #498, opened 2023-09-19
Avoid copying the contents of the `Common` project into all dependent subprojects.


orendain:
Currently, in order to avoid creating another assembly (DLL) that needs to be including with installs, we effectively copy the contents of the shared `Common/` project into every other mod project that needs it.  That way. there is no `Common.UI.dll`, and the necessary types are embedded as resources.

However, this approach stops `Common/UI` from being used as true library.  Access modifiers are set based on the fact that `Common/*` files are considered internal to other projects (because the comon files are effectively copy/pasted in) and don't truly reflect correct library access levels.  Additionally, builds take longer and there is a greater risk of namespace conflicts.

Instead, we ought to:
1. _Reference_ the `Common` project from other mod projects.  At runtime, the `Common` assembly then needs to exist.  We have a choice of either publishing a `Common.UI.dll` assembly that all must download, or simply stitch it to the resulting mod DLL file.  The latter would be better.
2. Separate the monolithic `Common` project into minilibraries.  At the moment, there is only one, so renaming to `Common.UI` will suffice.
3. To accomplish the last part of the first (1.) point, we need to clean up the referenced assemblies each mod relies on, and stitch all dependent libs into the single resulting mod assembly (DLL).

PR #499, opened 2023-09-19
Seperate monolithic `Common` project into individual subprojects.

orendain:
Part of https://github.com/orendain/DemeoMods/issues/498.

Using "monolithic" in the title may be misleading, as there is only a single common component: `UI`.

So really, this is more of a rename of `Common` -> `Common.UI`, so that other `Common.<something>` projects can come along.

Rather than use underscores to separate name parts (e.g., `Common_UI`), this PR opts for dots which is the .NET convention.

This PR also deletes `Common/CommonModule.cs`, which has gone unused.


TheGrayAlien (approved):
It works

PR #500, opened 2023-09-19
Adjust privacy modifiers to allow the common library to be used externally.

orendain:
Part of https://github.com/orendain/DemeoMods/issues/498.

In order to have the `Common.UI` subproject be reused as a library (externally, and not have to force what is [effectively a copy/paste during compilation](https://github.com/orendain/DemeoMods/blob/dd74d7c84b69e5fde529d4ab6ace201c457d3d83/HouseRules_Configuration/HouseRules_Configuration.csproj#L76-L78)), some access modifiers need adjusting to expose classes/methods.


TheGrayAlien (approved):
Looks like still a WIP, but still builds for me after last PR was approved

PR #503, opened 2023-09-21
Draw line-of-sight highlights on a different layer than player move highlights.

orendain:
Draws line-of-sight highlights on a separate layer (`TileHighLightController.TileHighLight` vs `TileHighlightController.PlayerPieceTurnHighlight`) so as to have those two highlights coexist rather than drawing over each other.


TheGrayAlien (approved):
Thanks! Works much better this way

PR #504, opened 2023-09-21
Retarget documentation links to the newly-renamed HouseRules directories.

orendain:
Retarget documentation links in to the newly-renamed HouseRules directories.  This PR also deletes two files that were successfully copied over, but somehow were not removed from their original location.


TheGrayAlien (approved):
LGTM

PR #518, opened 2023-10-02
Update exported example ruleset files.

orendain:
Exported all of the built-in rulesets to files to refresh the examples directory.

@TheGrayAlien: What should we do with the Revolutions ruleset JSON file?  It won't load in its current condition, and not sure what the intended fix is.


jimconner (approved):
LGTM :-)

TheGrayAlien:
There shouldn't be a Revolutions JSON in the HouseRules build. You might have done leftovers from installing my Revolutions rebuild.

PR #539, opened 2024-03-13
Update README.md

TheGrayAlien:
Noted BepInEx for PC and VR and MelonLoader is PC only now


orendain (approved):
Woah, good to know!  Thanks!

PR #541, opened 2024-03-23
Fix for 3/21 Demeo game patch

TheGrayAlien:
So silly... they changed one single number in a method and it caused all this fuss with BepInEx.


orendain (approved):
Good catch!

PR #545, opened 2025-06-08
Update Highlighter to work with Demeo 1.39.

orendain:
Updates Highlighter to work with Demeo 1.39.

Additional changes:
- Renames some variables (e.g., `currentPiece`) to make it more clear what they're used for.
- Adds some better comments to code.
- Extract the ability which is used to compute the highlight distance rendered.

> [!NOTE]
> @orendain tested this on PC Edition, but has not yet tested it on Steam VR or Quest VR. If needed, he can test that platform after a few other changes are merged in.

> [!NOTE]
> @orendain will update the build CI in a subsequent PR so that the build does not consistently fails on GitHub. For now, this is built/tested locally.

PR #546, opened 2025-06-08
Increase line of sight distance rendered.

orendain:
> [!NOTE]
> Please review / merge https://github.com/orendain/DemeoMods/pull/545 first.

---

Increases line of sight distance rendered. Specifically, uses the `HunterArrow` to decide the distance rendered (which is longer than the previous `MagicMissle`).

> [!NOTE]
> @orendain will update the build CI in a subsequent PR so that the build does not consistently fails on GitHub. For now, this is built/tested locally.

### DIFF
--- a/Highlighter/MoveHighlighter.cs
+++ b/Highlighter/MoveHighlighter.cs
@@ -12,7 +12,9 @@
 
     public static class MoveHighlighter
     {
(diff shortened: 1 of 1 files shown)


TheGrayAlien (approved):
Good catch

PR #548, opened 2025-06-08
Update SkipIntro to work with Demeo 1.39.

orendain:
Updates SkipIntro to work with Demeo 1.39.

Interestingly, the existing SkipIntro DLL works when I tested it. However, the existing _code_ won't compile, it's just the current DLL that works. This change will ensure SkipIntro will build fine.

> [!NOTE]
> @orendain tested this on PC Edition, but has not yet tested it on Steam VR or Quest VR. If needed, he can test that platform after a few other changes are merged in.


TheGrayAlien (approved):
Looks good

PR #556, opened 2025-06-10
Oof...

TheGrayAlien:
Make House Rules menu visible on PC. ;)

I honestly might have never noticed the problem without Mex's RoomFinder fix, though...


orendain (approved):
LGTM!

PR #562, opened 2025-06-11
Replaced with Exterminate

TheGrayAlien:
because that was somehow a better name for HailOfArrows in the code...?


orendain (approved):
> because that was somehow a better name for HailOfArrows in the code...?

Hehehe

PR #565, opened 2025-06-11
Fix OneMoreThing and AdamantPotion

TheGrayAlien:
OneMoreThing is now FreeAP.
AdamantPotion is now Invulnerability.
Replaced so the default rulesets will load!


orendain (approved):
LGTM

PR #567, opened 2025-06-11
Target zero seems to cause it

TheGrayAlien:
Maybe change it to 1? Just took it out for now to let the ruleset work.


orendain (approved):
LGTM

PR #572, opened 2025-06-15
Remove scorpions

TheGrayAlien:
In testing if attacking a scorpion that comes from a sandpile (or if they bury themselves) it causes the game to freeze for everyone. Has something to do with the Emerge behaviour. Even if you remove the behaviour it can still happen. Better safe than sorry.


orendain (approved):
How odd! Better safe than sorry, indeed.

PR #575, opened 2025-06-18
Fix multiple, overlapping, instances HouseRules UI appearing.

orendain:
Issue: When changing pages in the HR UI, multiple pages are overlapping onto one another.

Root cause: The process that determines whether or not to load UIs (our `OnSceneLoaded()` hook function) is triggering the HR UI be loaded multiple times.

Solution: Our `OnSceneLoaded()` is re-written to be more explicit about how to determine when to load in the UI. It follows the same approach as https://github.com/orendain/DemeoMods/pull/549: to be very explicit about which scenes, and on which platforms, constitute being in the "lobby" where we want to UI rendered.

---

Additional background: Some time back, RG added helpers as part of their `MotherbrainGlobalVars` class, capturing the exact platform the game is being run on. This didn't exist before, and so at the time we introduced a helper (`Common.UI/Environments`) to help us determine that ourselves. Our method has become a bit flaky as the game as been ported to different platforms and seen changes in scene numbers. It is now uneeded as RG has captured that info on their end, which we can piggyback on to determine the platform and scene more precisely.

Hence the removal of some of the code in `Common.UI/Environments`.

### DIFF
--- a/Common.UI/Environments.cs
+++ b/Common.UI/Environments.cs
@@ -2,42 +2,20 @@
 {
     using UnityEngine.SceneManagement;
 
-    public enum Environment
-    {
-        Hangouts,
-        NonVr,
-        Vr,
-    }
-
     public static class Environments
     {
-        private const int SteamHangoutsSceneIndex = 45;
-        private const int OculusHangoutsSceneIndex = 43;
-
-        public static Environment CurrentEnvironment()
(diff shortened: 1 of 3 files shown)


TheGrayAlien (approved):
Great job! Working on all platforms after testing!

PR #577, opened 2025-06-21
New rule PieceMagicStatAdded

TheGrayAlien:
Adds points to magic stat to defined heroes and increases their max allowed magic stat by the same amount. This can't be accomplished in PieceConfigAdjusted as far as I can tell.
Example:
    {
      "Rule": "PieceMagicStatAdded",
      "Config": {
        "HeroSorcerer": 1
      }
    }


orendain (approved):
Good stuff!

